PR #23, opened 2016-02-27
Preserve numpy array memory alignment flag

### LINKED ISSUE #22: zmq_stream producing "unaligned" np.ndarrays
A zmq_stream produces "unaligned" np.ndarrays, which theano doesn't know how to deal with. You can force a copy with np.array(), but that seems to defeat the purpose of the copy=False mode, no?

To Reproduce:
```
import pescador
import numpy as np
import theano
import theano.tensor as T

# A theano function that does something with two matrices
def theano_function():
    x = T.matrix('x')
    w = T.matrix('w')
    return theano.function([x, w], T.mean(T.dot(x, w)))

# A ZMQ Stream for testing
def data_gen():
    while True:
         yield {'X': np.random.random([8, 8])}
stream = pescador.Streamer(data_gen)
zmq_stream = pescador.zmq_stream(stream)

W = np.random.random([8, 8])
f = theano_function()

# This proves that the result is unaligned:
print (next(zmq_stream)['X'].flags.aligned)  # >>> False

# This call will fail.
f(next(zmq_stream)['X'], W)

# But this succeeds, because the copy forces the alignment
f(np.array(next(zmq_stream)['X']), W)

print (np.array(next(zmq_stream)['X']).flags.aligned)  # >>> True
```

This is the error I get:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-17de2968d2f1> in <module>()
----> 1 f(next(zmq_stream)['X'], W)

/usr/local/museami-dev/Theano-0.7.0-py2.7.egg/theano/compile/function_module.pyc in __call__(self, *args, **kwargs)
    511                     try:
    512                         s.storage[0] = s.type.filter(arg, strict=s.strict,
--> 513                                 allow_downcast=s.allow_downcast)
    514 
    515                     except Exception, e:

/usr/local/museami-dev/Theano-0.7.0-py2.7.egg/theano/tensor/type.pyc in filter(self, data, strict, allow_downcast)
    178                             "object shape", data.shape,
    179                             "object strides", data.strides,
--> 180                             "object dtype", data.dtype)
    181 
    182         i = 0

TypeError: ('Bad input argument to theano function with name "<ipython-input-2-96a0c586bb31>:5"  at index 0(0-based)', 'The numpy.ndarray object is not aligned. Theano C code does not support that.', 'object buffer\xb01\x1fHR\x90\xae?\x80446LG\xbd?\x04h "\x87\xfb\xce?\
.... [a bunch of crap here] ..... x1b\xcc\xed?', 'object shape', (8, 8), 'object strides', (64, 8), 'object dtype', dtype('float64'))
```

bmcfee:
This should fix #22

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,7 @@ python:
 
 before_install:
     - bash .travis_dependencies.sh
-    - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-        export PATH="$HOME/env/miniconda27/bin:$PATH";
-      elif [[ "$TRAVIS_PYTHON_VERSION" == "3.4" ]]; then
-        export PATH="$HOME/env/miniconda34/bin:$PATH";
-      else
-        export PATH="$HOME/env/miniconda35/bin:$PATH";
-      fi
+    - export PATH="$HOME/env/miniconda$TRAVIS_PYTHON_VERSION/bin:$PATH";
     - hash -r
     - source activate test-environment
 
--- a/.travis_dependencies.sh
+++ b/.travis_dependencies.sh
@@ -13,9 +13,10 @@ conda_create ()
     conda info -a
     deps='pip numpy scipy nose pyzmq scikit-learn'
 
-    conda create -q -n $ENV_NAME "python=$1" $deps
+    conda create -q -n $ENV_NAME "python=$TRAVIS_PYTHON_VERSION" $deps
 }
 
+src="$HOME/env/miniconda$TRAVIS_PYTHON_VERSION"
 if [ ! -f "$HOME/env/miniconda.sh" ]; then
     mkdir -p $HOME/env
     pushd $HOME/env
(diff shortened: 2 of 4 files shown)


bmcfee:
The plot thickens... This seems to work on Python 3 but not on 2.

bmcfee:
I think this actually comes down to using `memoryview` (python 3) vs `buffer` (python 2).  The former preserves alignment, the latter apparently does not.

This might mean that the desired behavior will only be possible in python 3, and we have to drop legacy python support.

I'm kind of okay with this.  What do you think, @ejhumphrey @cjacoby?

cjacoby:
I think as long as it's clearly documented.

That means only the ZMQ streamer part of Pescador will be limited to
python2 right? The rest of it should still work fine.

bmcfee:
Limited to Python 3, yeah.

So we have two options then:

1. Raise an exception of using zmq stream on py2
2. Let the user do it, but issue a warning.

Thoughts?

bmcfee:
I implemented option 2 (issue alignment warning on legacy python).  I think this one's set.

cjacoby:
I dig; LGTM!

Issue #31, opened 2016-08-01
Buffer Streamer as a subclass of Streamer


cjacoby:
Proposed refactor / feature / update / future release:
* Turn the `buffer_streamer` function into a `BufferStreamer` class, as a subclass of `pescador.Streamer`. It's `generate()` function would then yield batches of `batch_size`.
* Optionally: `ZMQStreamer` also being a subclass of `pescador.Streamer`

This makes a lot of intuitive sense to me, and would allow introspection of a sort on the "Buffer Streamer". Under OO semantics, I think that `BufferStreamer` logically "is a" `Streamer`, and `ZMQStreamer` also "is a" `Streamer`.

Thoughts / reasons not to do this? (other than time/effort; I can probably help with that, although  maybe not until I've finished a couple of other things).

bmcfee:
I think you're right.

If I had to take a stab at explaining why `zmq_streamer` and `buffer_streamer` are functions and not classes, it's that I wanted to keep `Streamer` dead simple, and use `has-a` semantics on buffering and IPC, rather than `is-a`.  This way, you could decouple the streamer construction from how it gets passed around and ultimately consumed.

I suppose you could always stick an upcast/`from_streamer` class method on the derived classes for the same effect.

I like promoting `ZMQStreamer` to a class -- it would make it possible to easily restart IPC streamers, say, across iterations of a cross-validation loop.

bmcfee:
While we're at it: what do you think about promoting `mux` into a streamer?

Its pattern is different from the others, but it still exposes the same consumption/`generate` interface.

cjacoby:
I was thinking about it, but I hadn't made up my mind on that yet, so I didn't mention it above.

My current thinking is that a "Streamer" minimally:
* consumes another generator(s?) of some kind.
* has a `generate()` function to expose that generator
* Maybe has a `next()` and/or `__iter__()` which wraps the `generate()` in a nice way.

I think by that definition, then a mux is just a streamer that takes multiple streamers as input, so I would agree, but then results in a streamer. (This would actually make it so you don't need to often wrap a mux in a streamer, which would be nice.)

bmcfee:
> This would actually make it so you don't need to often wrap a mux in a streamer, which would be nice.

Exactly.


> has a generate() function to expose that generator

For the purposes of an API, this is all that really matters.


> Maybe has a next() and/or __iter__() which wraps the generate() in a nice way.

[...]

PR #34, opened 2016-08-17
Convert mux, buffer_streamer, zmq_stream to Streamer subclasses

cjacoby:
This PR implements #31. I have for the most part tried to keep the changes as minimal as possible, optimizing for keeping the changes minimal and easily testable with the existing test suite. As such there are definitely some components which could still be improved (in terms of maximum OO shinyness), but this is a good first pass, I think.

* `Streamer` is now a parent class for all Streamers. The interface is now consistent between all streamers.
    * A `Streamer` can consume another `Streamer`, a generator, or a callable which produces a generator. (One exception is Mux, which can consume a list of streamers)
    * A `Streamer` produces samples from the `generate()` function.

* The base `Streamer` class now lives in `pescador.core`
* `utils.mux` is now a `Streamer`: `pescador.mux.Mux`
* `utils.buffer_streamer` is now a `Streamer`: `pescador.buffered.BufferedStreamer`
* `utils.zmq_stream` is now a `Streamer`: `pescador.zmq_stream.ZMQStreamer`

---
This change is [Reviewable](https://reviewable.io/reviews/bmcfee/pescador/34)


cjacoby:
@bmcfee This PR breaks backwards compatibility; is that a problem? Or should there be temporary wrappers in utils for these classes which raise deprecation warnings?

bmcfee:
I have no problem with breaking api for this change. We can call it 1.0 series of you like. Will review later today.

bmcfee:
Reviewed 2 of 13 files at r1.
Review status: 2 of 12 files reviewed at latest revision, 8 unresolved discussions.

---

*[pescador/core.py, line 3 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNiC1uqnPrVbpjmcc_:-KPNiC1uqnPrVbpjmcca:b-ac755g) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L3)):*
> ```Python
> #!/usr/bin/python
> """Core classes"""
> from __future__ import print_function
> ```

do we use print anywhere?  or is this a holdover from debugging?

---

*[pescador/core.py, line 31 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNiHBOeHXGJ4WE6UCh:-KPNiHBOeHXGJ4WE6UCi:bf5nyxj) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L31)):*
> ```Python
>     generator : iterable
>         A generator function or iterable collection to draw from.
>         May be another instance or subclass of Streamer.
> ```

I'd revise the type description to "iterable or Streamer"

---

*[pescador/core.py, line 76 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNiR9ILQVzc2Skg7gX:-KPNiR9ILQVzc2Skg7gY:b-yuy646) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L76)):*
> ```Python
>         if not (inspect.isgeneratorfunction(streamer) or
>                 isinstance(streamer, collections.Iterable) or
>                 isinstance(streamer, Streamer)):
> ```

pretty sure you can tuple-ize this:  `isinstance(streamer, (collections.Iterable, Streamer))`

---

*[pescador/core.py, line 85 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNik5hkOpUhzcc6f2D:-KPNik5hkOpUhzcc6f2E:b-3x87o3) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L85)):*
> ```Python
>         self.stream_ = None
> 
>     def is_active(self):
> ```

should we make this a `@property`?

---

*[pescador/core.py, line 135 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNjEHBrHwU63aKlUsu:-KPNjEHBrHwU63aKlUsv:b-axeoe1) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L135)):*
> ```Python
> 
>         # Resets the streamer so that we can restart it as necessary.
>         self.close()
> ```

This part makes me nervous; will it run even if `self.stream_` raises an exception?

It might be worth wrapping the setup (`activate`) and tear-down (`close`) inside a context manager.

---

*[pescador/core.py, line 150 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNjWXJ2V9UaE73m72P:-KPNjWXJ2V9UaE73m72Q:bx94o5r) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L150)):*
> ```Python
>         # ??? What more does this need?
>         while True:
>             yield self.generate()
> ```

I don't think this works?  In 3.5 you can do `yield from self.generate` but in general, i think it needs to be
```python
for item in self.generate():
    yield item
```

But maybe I'm missing something?

---

*[pescador/version.py, line 5 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNjqZP3YXTbxXsrajw:-KPNjqZP3YXTbxXsrajx:b8qk19n) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/version.py#L5)):*
> ```Python
> """Version info"""
> 
> short_version = '0.2'
> ```

please hold version updates until a release PR.

(I know, I meant to add this to the contributing doc which we don't have... :))

---

*[pescador/zmq_stream.py, line 29 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNjxAGdjxF7LHmDT2C:-KPNjxAGdjxF7LHmDT2D:b-vk455s) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/zmq_stream.py#L29)):*
> ```Python
>     from joblib._parallel_backends import SafeFunction
> 
> import pescador.core
> ```

Needs to be `from . import core` for proper path encapsulation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bmcfee/pescador/34)*

bmcfee:
Reviewed 4 of 13 files at r1, 6 of 7 files at r2.
Review status: 10 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[pescador/core.py, line 85 \[r1\]](https://reviewable.io:443/reviews/bmcfee/pescador/34#-KPNik5hkOpUhzcc6f2D:-KPSy_3hXBFe-HleCkeD:bqezvqz) ([raw file](https://github.com/bmcfee/pescador/blob/3d32e9b3bc00628ccb47f65c0212c425d60e1e19/pescador/core.py#L85)):*
<details><summary><i>Previously, bmcfee (Brian McFee) wrote…</i></summary>
> should we make this a `@property`?
</details>
I suppose I should have been more clear about the property usage.

1. If  `active` is a property, then we can just check `streamer.active`
2. We can use a setter to handle (de)activation.  This would only really work if the `activate` method has no parameters -- I'm not sure why it has `max_batches`?  This seems redundant with the argument in `generate`.

Given the above, we could then do `streamer.active = True` to activate and `streamer.active = False` to deactivate; the setter would handle the logic given the provided value.

Maybe this is less clean than `streamer.activate()`, `streamer.deactivate()`, and `streamer.is_active`?  I'm not sure yet, but it was just a thought.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bmcfee/pescador/34)*

cjacoby:
You're right; that `max_batches=None` is copypasta/garbage. Fixed.

I think the function approach is more clear by far, from a enduser-standpoint.

I also addressed the extraneous `is`, which was more or less an oversight.

bmcfee:
Reviewed 1 of 13 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bmcfee/pescador/34)*

bmcfee:
Is there anything left for this one?  Or should we merge?

cjacoby:
I think we're good.

... Not to say there isn't more to do, but I think everything in the scope
of this pr is good.

bmcfee:
Alright. Thanks!

Issue #21, opened 2016-02-25
[Proposal] Make the sklearn dependency soft / optional


ejhumphrey:
The guts of pescador are pure python (fishsnake), with two exceptions:
- pyzmq for background threading
- sklearn for BaseEstimator inheritance

The former is kind of core to what pescador is, whereas the second is just for maintaining a common interface with another library (which you could very well never touch). Worth noting, sklearn, and thus scipy, are bloating this build on travis, and any project that touches pescador. 

Granted there's a decent chance that anyone using pescador is probably going to be using one or both of those libs, but it's kinda bumming me out that pescador build / install times are almost double digit minutes, while the tests themselves take seconds.

discuss.

bmcfee:
Yeah, that's reasonable, but it will take some api breakage to do it properly.  (Moving StreamLearner into its own submodule.)

ejhumphrey:
What's the timeline for v0.2?

bmcfee:
None as of yet, since I haven't seen much need for major revision until now.

bmcfee:
Because I'm a lazy individual, #23 also implements a modification to the travis configs that should bring the build time back down to manageable.

bmcfee:
Alternately: how do you feel about just dropping the streamlearner object?  I don't think it gets used much (ever?) and dklearn does a reasonable enough job of this sort of thing anyway for people that just want out-of-core learning.

cjacoby:
I have a new repo of basically "pescador tools" that I could wrap it into. It might make sense there. Otherwise, I think it's probably fine to drop it.

https://github.com/cjacoby/streambuilder
It's only a few days old, and I'm still in the middle of setting up a bunch of things, but you should get the idea. (The current readme is a tad out of date, but at least explains what I'm getting at.) I found myself building this thing at least twice, so I decided to make a repo of it.

PR #35, opened 2016-08-22
removed StreamLearner class

bmcfee:
implements #21 by removing streamlearner altogether.

---
This change is [Reviewable](https://reviewable.io/reviews/bmcfee/pescador/35)


cjacoby:
LGTM.

An idea (perhaps an issue?): perhaps pescador should become an "organization", and there should be another repo with "recipes" or "helpers"? I have a couple of Pescador-wrappers where I'm trying to encapsulate primary use-cases of Pescador nicely, and I think that this StreamLearner basically belongs in that collection of things.

Thoughts? Although, probably we could move this discussion to an issue/question, since this PR is donezo.

PR #38, opened 2016-09-19
Develop/0.2

bmcfee:
Implements #31 

- [x] Doc/example updates for new API


bmcfee:
Docs are a must; backwards compatibility layer is not so important IMO; that's what major revisions are for.  I'm okay with bumping this up to 1.0 instead of 0.2.

cjacoby:
I think if we're going to call it 1.0, then I might want to be more
thoughtful about things it needs to include, so would rather do 0.2 for
now, and use it for a little bit to see if anything else comes up.

bmcfee:
> I think if we're going to call it 1.0, then I might want to be more
> thoughtful about things it needs to include, so would rather do 0.2 for
> now, and use it for a little bit to see if anything else comes up.

Not that this would be the entire PR for 1.0; just that the changes here are sufficient to put us in major revision territory, so we should change the version accordingly.

bmcfee:
Hey @cjacoby @ejhumphrey I took a quick stroll through this PR, cleaned up some docs and build issues, and I think it's good enough to merge.

Any opposition to abandoning 0.2 as a milestone and pushing for 1.0?

Issue #41, opened 2016-11-08
Leaky Numpy (NPZ) Gotchas


ejhumphrey:
First off, this is not strictly a pescador issue / bug. **However**, given the normal use cases of pescador, I do think this has the potential to be a common trap, and is worth documenting for posterity (especially here, which is where I assume future inquiries might land first). 

As I've tried to explore and document in [this repository](https://github.com/ejhumphrey/pescador-wtf), there are **two** things one must do to avoid memory leaks when pulling data from numpy's archive files ('.npz').

- Use a ContextManager when accessing the file, e.g. `with np.load(npz_file) as data: ...`
- Do not yield references to ndarrays via `dict`s; arrays created in the yield statement are kosher  e.g. `yield dict(x=np.array(x[n:n+k]))`

Both of these points make perfect sense in hindsight, though neither were particularly obvious to me at first. Am I alone in my oversight? What's the best way to document this?

### Appendix:
For clarity, a safe NPZ archive sampler looks like the following:
```
import numpy as np

def npz_sampler(npz_file):
    """Randomly sample aligned rows from an NPZ archive"""
    with np.load(npz_file) as data:
        x, y = data['x'], data['y']

    while True:
        i = np.random.randint(len(x))
        yield dict(x=np.array(x[i:i + 1]), y=np.array(y[i:i + 1]))
```

ejhumphrey:
aaand way too many hours later, @cjacoby I think you'll be most interested in this.

cjacoby:
I'm not sure I follow the second bullet; can you provide an example of what *not* to do?

I am probably doing all of these things, so I am extremely grateful for the pointers.

IMO there should be a separate pescadores repo (I've said this elsewhere / I have the beginnings of this already) with helper functions for extremely common things like this.

ejhumphrey:
both of these are bad:

```
def leaky_sampler_a(filename):
    with np.load(filename) as data:
        x = data['x']

    while True:
        i = np.random.randint(len(x))
        yield dict(x=x[i:i + 1])


def leaky_sampler_b(filename):
    x = np.array(np.load(filename)['x'])
    while True:
        i = np.random.randint(len(x))
        yield dict(x=x[i:i + 1])
```
you see what's going on? 

[...]

Issue #40, opened 2016-11-08
canonical examples


ejhumphrey:
Based on my experience / sufferings, I think it'd be nice to have some canonical examples of The Right Ways to use pescador as benchmarkable scripts (timings and throughput). A few come to mind...

- [x] reading from NPZ files
- [x] data augmentation
- [x] parallelism (updated)

any others?

cjacoby:
:+1:

We could make another repo in the pescadores/ with examples, or something
like that.

Some of those things may end up being so common as to want some kind of
wrapper library? I've found this to be true, in practice.

bmcfee:
Implemented as #47

PR #47, opened 2016-12-26
More advanced and thorough examples in docs

bmcfee:
This PR implements #41 #40 -- adding examples for handling NPZ files and parallelism.


bmcfee:
Probably a good idea to include a fully worked example in the docs, eh?  Eg data augmentation on iris via sklearn.

bmcfee:
Tagging @justinsalamon

docs/example2.rst
```diff
@@ -48,28 +48,49 @@ As a concrete example, we can simulate a mixture of noisy streams with differing
...
 
         # Instantiate a linear classifier
         estimator = SGDClassifier()
 
         # Build a collection of Streamers with different noise scales
         streams = [Streamer(noisy_samples, X[train], Y[train], sigma=sigma)
-                   for sigma in [0.5, 1.0, 2.0, 4.0]]
+                   for sigma in [0, 0.5, 1.0, 2.0, 4.0]]
 
         # Build a mux stream, keeping only 2 streams alive at once
```
cjacoby: Here it says 2 streams, below 3. I assume this is not what you mean?

docs/example2.rst
```diff
@@ -48,28 +48,49 @@ As a concrete example, we can simulate a mixture of noisy streams with differing
...
         mux_stream = Mux(streams,
-                         k=2,    # Keep 2 streams alive at once
-                         lam=16) # Use a poisson rate of 16
+                         k=3,    # Keep 3 streams alive at once
+                         lam=64) # Use a poisson rate of 64
```
cjacoby: In the below text, you say "...`lam=16`, but it's now 64. Plzfix.

bmcfee: Done.

justinsalamon:
@bmcfee happy to pitch in once im back, if needed.

docs/example3.rst
```diff
@@ -0,0 +1,57 @@
...
+        # so that memory is released when the streamer is deactivated
+        with np.load(npz_file) as data:
+            # How many rows are in the data?
+            # We assume that data['Y'] has the same length
+            n_total = len(data['X'])
+
+            while True:
+                # Compute the index offset
+                idx = np.random.randint(n_total - n)
+                yield dict(X=data['X'][idx:idx+n],
```
ejhumphrey: IIRC, this kind of slicing will (should?) cause memory leaks because the dictionary inherits the identifier of `data`, despite using the context manager. I believe [this demonstrates the behavior.](https://github.com/ejhumphrey/pescador-wtf/blob/master/test_memleak.py#L104)

I think the best practice here is to force the data copy before handing an ndarray to the dictionary, like `dict(X=np.array(data['X'][idx:idx + n]...`.

note: it makes me *wicked* uncomfortable how easy this is to overlook, but I don't have any answers I like. There are three that jump out at me:

- `Streamer` objects have a keyword argument like `asarray=True` which, by default, forces values in dictionaries to be reallocated as new arrays. I _think_ this would do the right thing, but I'm maybe 65% sure of that. Major win, the default behavior is sane; potential drawback, how detrimental to performance is it to blindly reallocate all matrix data?
- We provide an `@asarray` decorator, that one can apply to generators to enforce the data it yields is properly reallocated, allowing Python to fully clean up after a stream when it is deactivated. Major disadvantage here is that this should be the default behavior, but decorators are explicit add-ons, and like to be forgotten.
- We provide a `Sample` object, as a subclass of `dict` that does the right thing with the data it is handed. Again, not exactly default behavior, and is starting to feel slightly over-engineered.

any other ideas?

nit: there should be whitespace around the `+`s

bmcfee:
> IIRC, this kind of slicing will (should?) cause memory leaks because the dictionary inherits the identifier of data, despite using the context manager. I believe this demonstrates the behavior.

I'm not sure that's a leak exactly - the reference is still active, so the GC won't (and shouldn't) free it.

However, once the context manager exits AND the last reference to the array is lost, then the GC should collect it.  If it's not, that's a bug in python.

It's worth mentioning that a buffered streamer does force the copy, so that could be an extra sanity check.

(In case it's not obvious, I really don't like doing explicit copies just to avoid apparent GC weirdness.)

bmcfee:
> I have also thought about making a sample object (I've done this in some of my own work), so that additionally the Sample object could use __slots__ = ['X', 'Y'], and thereby save a lot of memory.

The downside of this is that you'd be stuck with only the two pre-determined variable names. I sometimes have streamers that generate many more variables (eg for multitask learning for different Y's, or cross-modal learning for different X's).

Would it be possible to do something that works like a named tuple (dynamically set, but fixed fields), that you could still use in the way you describe above?

docs/example3.rst
bmcfee: I've thought a bit more about this, and I think it's best that we not over-complicate things.  It's already complicated enough to use, and I don't like the idea of doing reallocation behind the scenes.

FWIW, this thread is helpful: https://mail.scipy.org/pipermail/numpy-discussion/2015-December/074376.html

TLDR: `npy` files support [mmap mode](https://docs.scipy.org/doc/numpy/reference/generated/numpy.load.html) while `npz` files do not.  If you care about memory leaking here, don't use `npz` files.  If you want memory mapping with structured data, use `pytables` instead.

I think the sane thing to do here is to drop this example, and note the memory mapping problem in the docs.

bmcfee:
I updated the npz/npy example again to discuss memory mapping issues, and while I was at it, fixed the build on RTD.  @ejhumphrey care to take another pass on this?

ejhumphrey:
looks great 🎉

PR #48, opened 2016-12-26
Fixing timeout bug

### LINKED ISSUE #46: py3.5 Heisenbug in zmq tests on travis
This has popping up recently on travis: tests pass consistently on py27 and 3.4, but fail on 3.5.

Here's a [log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/185856127/log.txt), relevant snippet:

```
======================================================================
FAIL: test_zmq_streams.test_zmq(True, 1)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/env/miniconda3.5/envs/test-environment/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/pescadores/pescador/tests/test_zmq_streams.py", line 21, in __test
    eq_(len(reference), len(query))
AssertionError: 200 != 170
```

This is happening within the timeout-enabled test for zmq streaming.  I think what's happening is that there's some unpredictable latency in kicking off new processes on travis.  Often if I restart the build, it will pass on the second try.

This points to some kind of a logic error in the subprocess handling of the zmq streamer, where the streamer is killed before all data is consumed.

bmcfee:
This hopefully resolves #46 by fixing the join-timeout on ZMQ worker threads.

### DIFF
--- a/pescador/zmq_stream.py
+++ b/pescador/zmq_stream.py
@@ -90,9 +90,9 @@ def zmq_worker(port, streamer, terminate, copy=False, max_batches=None):
     try:
         # Build the stream
         for batch in streamer.generate(max_batches=max_batches):
+            zmq_send_batch(socket, batch, copy=copy)
             if terminate.is_set():
                 break
-            zmq_send_batch(socket, batch, copy=copy)
 
     finally:
         # send an empty payload to kill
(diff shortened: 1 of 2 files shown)


bmcfee:
I don't think this fundamentally changes how the workers operate, but it does seem to fix the race condition bug on travis.  Anyone care to confirm?

cjacoby:
I will take a look later today or tomorrow.

How did you test, other than the unit test itself?

Issue #51, opened 2017-01-04
Simplify dependencies


bmcfee:
The current install dependencies look like:

```python
    install_requires=[
        'joblib>=0.9',
        'six',
        'pyzmq',
        'numpy',
        'scipy',
        'scikit-learn',
    ],
```

Of these, only the first four are necessary at this point.

I also think we should port over the `SafeFunction` code from `joblib` to eliminate that dependency.  This would reduce dependencies down to

```python
six
pyzmq
numpy
```

which, I think, is about as minimal as it can be for this.

bmcfee:
Merged as #52

PR #52, opened 2017-01-04
removed unused dependencies

bmcfee:
implements #51


bmcfee:
Note: I opted to keep the joblib dependency.  It's pure python (low-overhead for install) and actually does a lot of non-trivial things for us in terms of exception handling.

Issue #43, opened 2016-12-21
Rewrite tests in py.test


bmcfee:
Since nose is abandonware

cjacoby:
I will try to do this in the next few days, while i'm not working. 👍 for py.test!

PR #55, opened 2017-02-27
Convert tests to pytest

cjacoby:
Adresses #43.

I tried to keep the test conversion simple, such that I use the `py.test` paradigm as much as possible, while keeping the intent of the original test.


bmcfee:
Thanks for taking this one! I'll look at it tomorrow.

tests/test_buffered_streams.py
```diff
@@ -1,75 +1,73 @@
...
 
+@pytest.mark.parametrize('n1', [5, 10, 15])
+@pytest.mark.parametrize('n2', [5, 10, 15])
+def test_batch_length__not_equal(n1, n2):
+    if n1 != n2:
+        generator, n = T.__zip_generator(3, n1, n2), n1
+
+        with pytest.raises(pescador.PescadorError):
+            for batch in generator:
+                assert pescador.buffered.batch_length(batch) == n
```
bmcfee: This test might be simpler if you tuple-ize (n1, n2) as parameters, and then mark out a couple of specific cases `(5, 10)` as failures.  [Here](https://github.com/bmcfee/pumpp/blob/master/tests/test_task.py#L419-L423) is an example illustrating how to do this.

bmcfee:
Hot diggity -- this got coverage working in the zmq parallel process?

bmcfee:
Also, I cleared the dependency cache and restarted the build.  Hopefully that will clear up the push failure.

bmcfee:
@cjacoby is there anything else you want to add here, or is it good to go?

cjacoby:
I think it's good to go. Yay for py.test!

Issue #56, opened 2017-03-02
Keras generator interface


bmcfee:
I've been using pescador with keras lately, and find myself writing some glue code to convert the pescador stream outputs (dict) into a keras-friendly format (tuple).

What do people think about adding a simple streamer method that selects out target entries from the dict? Something like the following:

```python
class Streamer:
    ...
    def tup(self, *items, **kwargs):
        for data in self.generate(**kwargs):
            yield (data[item] for item in items)
```
which you could then use as

```python
model.fit_generator(my_stream.tup('name_of_input_variable', 'name_of_output_variable'), ...)
```

ejhumphrey:
I'd rather fix keras than break pescador 😄 

but srsly, that's fine / makes sense.

but even more srsly, tuples are a bad idea.

bmcfee:
> but even more srsly, tuples are a bad idea.

been reading https://attrs.readthedocs.io/en/stable/why.html#tuples eh?

ejhumphrey:
ha, no, I think it's just common sense ... but thanks for arming me with a reference that I can now wield at people.

bmcfee:
anyway --
what should we call such a wrapper function?

cjacoby:
I was actually going to make this proposal myself, for the same reason.

`Streamer.tuples()` is okay with me.

One possible solution: make it a subclass of `Streamer`, like
`class TupleStreamer(Streamer)`, which returns tuples instead of dicts.

Then, `Streamer.tuples()` would actually do:
```
def tuples(self, *items, **kwargs):
    return TupleStreamer(self, *items, **kwargs)
```

Not 100% sure about it, but I currently think that might be a cleaner way. Will continue to mull over it.

bmcfee:
I think I prefer the original version, since any `Streamer` subclass would automatically inherit it as a wrapper over the `generate` method, so it would work seamlessly with mux, zmq, and buffer.

What's gained by having a separate class for tuple streams?

cjacoby:
I see your point, and agree 100%. Nevermind on that.

PR #57, opened 2017-03-02
tuple generator

bmcfee:
This implements #56 , and provides a tuple generator interface for all streamer objects.

To-do:

- [x] core implementation
- [x] core unit test for streamer
- [x] integration tests for all derived classes
- [x] cyclic tuple support

### DIFF
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -155,7 +155,7 @@ class Streamer(object):
             for item in self.generate():
                 yield item
 
-    def tuples(self, *items, max_batches=None):
+    def tuples(self, *items, **kwargs):
         '''Generate data in tuple-form instead of dicts.
 
         This is useful for interfacing with Keras's generator system,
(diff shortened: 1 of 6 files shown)


bmcfee:
I added cycle support as an optional kwarg to `tuples`.  Maybe this would be better as a separate method, but it seems kinda redundant.  What do yall think?

bmcfee:
I think this one's good to go -- tuple testing is integrated across all subclasses.

bmcfee:
.. getting a strange heisenbug on zmq streamer though.  Possibly timeout related.  @cjacoby @ejhumphrey want to take a look?

bmcfee:
Hrm.  Heisenbug averted?  This might take a more careful look.

For now though, I think ZMQ issues are safely independent of this PR, so it should be ready for review (and merge, I think).

bmcfee:
> Could try to go deeper.

Probably not worth it -- the changes are relatively superficial.

If there's anything to think about, maybe it's whether this is really the interface we want.  I think it is, but I'm open to suggestions.

cjacoby:
I'm okay merging it here, and changing it in the next couple days if one of
us changes our mind.

Right now I'm feeling good about this interface, though.

PR #69, opened 2017-03-18
Docs 1.0.0

bmcfee:
And general preparation for the 1.0 release

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,6 +1,39 @@
 Changes
 =======
 
+v1.0.0
+------
+
+
+- `#23`_ Preserve memory alignment of numpy arrays across ZMQ streams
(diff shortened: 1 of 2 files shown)


bmcfee:
Docs build and all links check out; I think this is good to go and then we can :shipit:

cjacoby:
Looking from my phone, but LGTM.

cjacoby:
One thought: Is it worth being more explicit about the break with the
previous API? This might be enough. Not sure.

bmcfee:
> Is it worth being more explicit about the break with the
> previous API?

Yup, done.  Not sure we have enough of a user-base that it matters that much, but I threw a line in the changelog about it.

cjacoby:
I can think of at least three other people it might effect ;).

Thanks :+1:

Issue #64, opened 2017-03-18
Simplified __call__ interface?


bmcfee:
I've been using pescador a bit over the last few weeks, and find myself wondering if we could tighten up the API a little before the 1.0 release.

Specifically, `streamer.generate()` is a bit verbose; how about adding a `__call__` method so we can instead do something like:

```python
stream = pescador.Streamer(my_generator, my_params...)

for x in stream():
    process(x)
```

This wouldn't help for `cycle` or `tuples`, but it would cut down a bit of boilerplate code for the most common use cases.

cjacoby:
I've also been using it a lot recently, and I like. 👍 

What if we kept the `generator()` interface internally, and did something like this?

```python
def __call__(self, cycle=False, **kwargs):
    if cycle:
        return self.cycle(**kwargs)
    else:
        return self.generate(**kwargs)
```

bmcfee:
I dig it.  Tuples are still an oddball, and I'd prefer to keep that one separate.

How about I implement this after merging the doc updates so that everything can be updated in one pass?

cjacoby:
Roger roger

ejhumphrey:
I realize I slept on this issue the first time around and apologize in advance, but I would be keen to revisit this discussion maybe? 😞  🙏 

Specifically, I'd propose making `core.Streamer` directly iterable, e.g. `[x for x in stream]` is a thing. 

*Commentary* 
- This isn't (necessarily) mutually exclusive with `[x for x in stream(max_batches=10, cycle=True)]` ... though it does violate the "There should be one good way to do a thing" principle. 
- My most common use case is consistent with the default args (which leads me to think that it is _the_ common use case) ... which means everyone using pescador will be writing `for stuff in stream()` all over the place, and those bonus parens smell like a minor anti-pattern. 
- Can the `cycle` and `max_batches` (or `max_samples`, re: #75) go in the init method? I don't immediately see a reason these would change over the lifetime of a Streamer, but am happy to be educated.

bmcfee:
> Can the cycle and max_batches (or max_samples, re: #75) go in the init method? I don't immediately see a reason these would change over the lifetime of a Streamer, but am happy to be educated.

I think this might cause some difficulties with how streamers get used inside `Mux`.  If you're using poisson-distributed sample limits, you would have a different `max_batches` each time the streamer is activated.  Ditto for `cycle`, though it's a bit less obvious there.

[...]

Issue #79, opened 2017-04-09
Mux a Mux stream


faroit:
For non obvious reasons I would like to multiplex a stream with an already multiplexed stream. 

```python
import pescador

def data_high():
    while True:
        yield dict(X=1)


def data_low():
    while True:
        yield dict(X=0)


def data_neg():
    while True:
        yield dict(X=-1)


high_stream = pescador.Streamer(data_high)
low_stream = pescador.Streamer(data_low)

mux_low_high = pescador.mux.Mux(
    [low_stream, high_stream], 2
)

mux_mux_neg = pescador.mux.Mux(
    [data_neg, mux_low_high], 2,
)

print([X for X in mux_mux_neg.generate(max_batches=10)])
```

Generating samples from `mux_mux_neg` results in an error:

```
Traceback (most recent call last):
  File "keras_example.py", line 130, in <module>
    print([X for X in mux_mux_neg.generate(max_batches=10)])
  File ".env/lib/python2.7/site-packages/pescador/mux.py", line 157, in generate
    with core.StreamActivator(self):
  File ".env/lib/python2.7/site-packages/pescador/core.py", line 14, in __enter__
    self.streamer.activate()
  File ".env/lib/python2.7/site-packages/pescador/mux.py", line 145, in activate
    self.__new_seed(self.stream_idxs_[idx]))
  File ".env/lib/python2.7/site-packages/pescador/mux.py", line 247, in __new_seed
    return (self.seed_pool[idx].generate(max_batches=n_stream),
AttributeError: 'function' object has no attribute 'generate'
```

Even though workarounds are simple and I can't think of any justfied applications where this would be more practical than just muxing the 3 streams, I just wonder if this is intended?

bmcfee:
The problem here is that `Mux` only accepts streamers, not arbitrary iterators.

If you replace `data_neg` with `pescador.Streamer(data_neg)`, it ought to work.

We should definitely have some type checking inside `Mux` to catch this kind of thing though.

bmcfee:
Side note: I know @ejhumphrey has been actively hacking away at making this whole process less cumbersome, he might have some insight to share on how this might change in the future.

ejhumphrey:
[...]

Issue #78, opened 2017-04-03
Refactoring tests?


ejhumphrey:
kay, last issue of the day before actually _doing_ something 😄 

I would like to indulge my OCD and refactor the tests directory to take a `{name}.py -> test_{name}.py` structure. By that paradigm, we're currently missing `test_core.py` and `test_buffered.py`... it's all getting tested, since our coverage is ≈97%, but I'm finding it less than obvious how / where to dive in and start making changes.

what say you @cjacoby @bmcfee?

cjacoby:
I am not offended by this, go for it as far as I'm concerned.

I have two bonus thoughts about this (these are on principle, requiring no actual changes to your proposals... yet)
1. I agree in general about `{name}.py -> test_{name}.py`, with the following modifications:
    * Due to the py.test inability to handle multiple test files with the same name, I usually do:
      `{submodule}/{name}.py => test_{submodule}_{name}.py`
    although that doesn't matter to us yet. (But would on your transforms proposal)

2. To me Integration/system tests or the like also don't necessarily follow this convention.

ejhumphrey:
closed with #88

Issue #77, opened 2017-04-03
Composing generators: good idea or great idea?


ejhumphrey:
Part of what I dig about pescador (and leveraging generators for data sampling) is that it's easy to build pipelines of well-encapsulated transforms. They could be used from anything from data augmentation, e.g. additive noise, to glue layers, e.g. "I need one-hot encoded vectors from integer indexes". While [pumpp](https://github.com/bmcfee/pumpp/) aims to achieve this, I imagine the pescador audience will be more broad, and some simple tools would be good to help outline (what we think are) good design practices.

For example, I've been thinking for sometime about having a `pescador.transforms` submodule which provides some decorators to turn sample-wise functions into iterable transforms, like...

```
def stateless(func):
     def wrapped(gen, *args, **kwargs):
         for samp in gen:
            yield func(samp, *args, **kwargs)
     return wrapped
```

This could allow for some easy fixes, like for #76 ...
```
@stateless
def newaxis(sample, *keys):
    for key in keys:
        sample[key] = sample.pop(key)[np.newaxis, ...]
    return sample

# Given a sample generator `stream`...
pescador.buffer_batch(newaxis(stream, 'x_in', 'y_true'), 10)
```

I'd want to disable the decorator for testing, but I'm not sure that's possible (and it's not the end of the world). This would also allow for some other reshaping / renaming that goes on regularly, such as transforming `dict`s to `tuple`s:

```
@stateless
def values(sample, *keys):
    return [sample[key] for key in keys]
```

thoughts? Is this a good idea?

cjacoby:
Generally 👍 . I have built things somewhat in this direction for work already, but this is slightly more elegant. 

I definitely agree that there are a number of `transforms` that are likely to be common use cases and useful.

bmcfee:
> thoughts? Is this a good idea?

This sounds slick.  I'm not 100% sure I understand your example though: is the idea to make a decorator that can wrap simple functions as generators?  Or as streamers?

ejhumphrey:
so here's a few thoughts...

after mulling over other issues (namely #75), I think transforms would process arbitrary data streams, i.e. an "online" pipeline. So, yea, this decorator would make a function iterable.

The decorator is one possible way of abstracting iteration from processing ... otherwise every transform is going to have the same `for x in stream: yield fx(x)` block. This could also be achieved with a class for stateful processing, or if you needed to have a transform upstream of a mux.

[...]

PR #80, opened 2017-04-09
v1.1 -> v2.0 Semantics refactor

ejhumphrey:
Renames tests for clearer mappings between submodules
Addresses much of the conversation from #75 
- will break the current API, and needs some help signaling deprecations
- `Streamer.generate` is now `Streamer.iterate`, since it returns an iterator
- `Streamer`types are now directly iterable
- `Mux` semantics are simplified to remove the (now superfluous) `seed` concept, which were synonymous with streams; now there are only streamers and streams.
- the notion of `batch` is scrubbed in all core functionality.

Will resolve: #64, #75, #76, #77, #78, #79, #81


pescador/buffered.py
```diff
@@ -56,31 +58,31 @@ def activate(self):
         """Activates the stream."""
         self.stream_ = self.streamer
 
-    def generate(self, max_batches=None):
+    def iterate(self, max_iter=None):
```
bmcfee: persist the old generate function here.  we can wrap the call signature for API compatibility, but there should be a warning.

pescador/buffered.py
```diff
@@ -56,31 +58,31 @@ def activate(self):
...
-                if max_batches is not None and n >= max_batches:
+                if max_iter is not None and n >= max_iter:
                     break
                 yield batch
 
 
-def buffer_batch(generator, buffer_size):
-    '''Buffer an iterable of batches into larger (or smaller) batches
+def buffer_batch(streamer, buffer_size):
+    '''Buffer data samples from an streamer into one data object.
```
bmcfee: this doesn't actually require a streamer, right?

pescador/core.py
```diff
@@ -8,6 +8,11 @@
 
 class StreamActivator(object):
     def __init__(self, streamer):
+        for mname in ['activate', 'deactivate']:
```
bmcfee: why not just check `isinstance(streamer, Streamer)` instead?

pescador/core.py
```diff
@@ -92,8 +98,8 @@ def __init__(self, streamer, *args, **kwargs):
 
         if not (inspect.isgeneratorfunction(streamer) or
                 isinstance(streamer, (collections.Iterable, Streamer))):
-            raise PescadorError('streamer must be a generator, iterable, or '
-                                'Streamer')
+            raise PescadorError('`streamer` must be a callable function that '
+                                'returns an iterable object.')
```
bmcfee: could we actually generalize this and support iterables as well as functions?  Call it a stretch goal.

pescador/core.py
```diff
@@ -113,57 +119,53 @@ def activate(self):
...
             # If it's iterable, use it directly.
-            self.stream_ = self.streamer
+            self.stream_ = iter(self.streamer)
 
     def deactivate(self):
         self.stream_ = None
 
-    def generate(self, max_batches=None):
-        '''Instantiate the generator
+    def iterate(self, max_iter=None):
```
bmcfee: again, deprecation rename / persist old signature with warning

pescador/core.py
```diff
@@ -113,57 +119,53 @@ def activate(self):
...
         while True:
-            for item in self.generate():
-                yield item
+            for obj in self:
+                yield obj
 
+    # TODO: ejhumphrey would like to deprecate this class method and turn it
+    #       into a "transform". @bmcfee please advise? example to follow?
     def tuples(self, *items, **kwargs):
         '''Generate data in tuple-form instead of dicts.
```
bmcfee: I'm still not 100% on what a `transform` is defined to be.

I find the `.tuples()` interface quite useful for interfacing with keras though, and I'd prefer to not have to replace it with something more cumbersome.

pescador/mux.py
```diff
@@ -19,24 +19,26 @@ class Mux(core.Streamer):
...
     ...     MY_FUNCTION(batch)
+
+    Mux([stream, range(8), stream2])
     '''
 
-    def __init__(self, seed_pool, k,
-                 lam=256.0, pool_weights=None, with_replacement=True,
-                 prune_empty_seeds=True, revive=False,
+    def __init__(self, streamers, k,
+                 lam=256.0, weights=None, with_replacement=True,
```
bmcfee: deprecation renames EVERYWHERE

while we're at it though, I'd like to replace `lam`=>`rate`

pescador/zmq_stream.py
```diff
@@ -81,15 +81,16 @@ def zmq_recv_batch(socket, flags=0, copy=True, track=False):
     return results
 
 
-def zmq_worker(port, streamer, terminate, copy=False, max_batches=None):
+def zmq_worker(port, streamer, terminate, copy=False, max_iter=None):
```
bmcfee: renames

bmcfee:
@ejhumphrey Re: handling variable rename deprecations, here's how librosa does it.  Note that the current stable version has no lingering renames, but rolling back to 0.4.3 you can find some examples.

1. Check out the `util.deprecation` [submodule](https://github.com/librosa/librosa/blob/446708889bf04438aaac7fe64aa738a09955bbf6/librosa/util/deprecation.py)
2. Here's an example of it in use: https://github.com/librosa/librosa/blob/446708889bf04438aaac7fe64aa738a09955bbf6/librosa/core/constantq.py#L327-L329
3. There's also some helpers in the `util.decorators` [submodule](https://github.com/librosa/librosa/blob/446708889bf04438aaac7fe64aa738a09955bbf6/librosa/util/decorators.py) for deprecating entire functions, or moving them to different submodules/renaming them.
4. Here's an example of move/rename behavior: https://github.com/librosa/librosa/blob/446708889bf04438aaac7fe64aa738a09955bbf6/librosa/core/audio.py#L692-L693
5. Here's an example of deprecated function behavior: https://github.com/librosa/librosa/blob/446708889bf04438aaac7fe64aa738a09955bbf6/librosa/feature/spectral.py#L1337-L1338

The [futurepast](https://github.com/amueller/futurepast) package implements some of these things and more, but it's not yet at an initial release.

pescador/buffered.py
ejhumphrey: roger that

pescador/buffered.py
ejhumphrey: no, it needs an iterable; not sure what a sane variable name would be here. I also aspire to change this on a subsequent PR, so .... any strong opinions?

pescador/core.py
ejhumphrey: **tl;dr:** ya, that's better.

**commentary:** I think I had a lapse of Java ptsd, and wanted to check that the object implements the interface of `activate`/`deactivate`, since it doesn't strictly matter that it's a Streamer (although when would we run into that? /shrug). If we really cared, a more Pythonic way of doing this would be to define an ActiveStream mixin, but ... this is total overkill for where we are right now, and mention it now only for completeness.

pescador/core.py
ejhumphrey: Actually it does as of [this change](https://github.com/pescadores/pescador/pull/80/files/44d82c225dd8ceb6777cead497a7682ee4eb195c#diff-535170d9182a6f744ecc1f98d69624d5R124) 😄  ... I've updated the error message, and added a `Streamer(iterable)` test in da42dbf1.

pescador/core.py
ejhumphrey: roger roger

pescador/core.py
ejhumphrey: I've added a [long-form response here](https://github.com/pescadores/pescador/issues/77#issuecomment-292806012), including why I think this is an anti-pattern.

**tl;dr:** `Streamer.tuples` only works for a subset of data streams that might be fed through it; I'd propose explicitly separating `Streamer` functionality from processing data streams. 

The thing I'd like to replace it with is a separate `tuples` function (as sketched out in that linked issue), which I'd argue isn't any more cumbersome than what we currently have and Streamer remains data stream agnostic.

pescador/mux.py
ejhumphrey: haha roger

pescador/zmq_stream.py
ejhumphrey: yayaya

pescador/buffered.py
ejhumphrey: @bmcfee do you have an example of "@move"ing a class's method? I could hack it out of if need be, but I'd rather do the fast thing.

pescador/buffered.py
ejhumphrey: I've opted for `stream`

ejhumphrey:
thanks @bmcfee, made it pretty far ... still have some work to do, but any advice / pointers will be helpful.

bmcfee:
@ejhumphrey I don't have an example of `@move`ing a class method, because that gets a bit more involved. However, I think you don't necessarily want or need that here, since the call signature has changed as well.  I would just write a shim function to translate to the new API, and a custom warning message.

cjacoby:
(I have made a first pass looking through this; I won't have time for another look until tomorrow, because work+taxes)

ejhumphrey:
guh, I need to get the test coverage back up (idk what's up with that), but this is in a state where it's worth taking another look @bmcfee ... I think I did most of the deprecation stuff correctly, but do let me know if that's not the case (I have no doubt you will :o)

pescador/maps.py
```diff
@@ -0,0 +1,120 @@
...
+    y : np.ndarray, list of np.ndarray, or None
+        If `outputs` is a string, `y` is a single np.ndarray.
+        If `outputs` is an iterable of strings, `y` is a list of np.ndarrays.
+        If `outputs` is a null type, `y` is None.
+    """
+    if inputs and isinstance(inputs, six.string_types):
+        inputs = [inputs]
+    if outputs and isinstance(outputs, six.string_types):
+        outputs = [outputs]
+
```
bmcfee: I don't think we should upcast to list here.  Keras's interface is overloaded to support tensors or lists of tensors as inputs (or outputs). The following are all perfectly valid:

```python
model = K.models.Model(x1, y1)
model = K.models.Model(x1, [y1, y2])
model = K.models.Model([x1, x2], [y1, y2])
```
where `x*` and `y*` are tensors.

`keras_tuples` should let me do any of the following:

```python
keras_tuples(streamer, inputs='X1', outputs='Y1')  # generates tuples of (ndarray, ndarray)
keras_tuples(streamer, inputs='X1', outputs=['Y1', 'Y2'])  # generates tuples of (ndarray, list of ndarray)
keras_tuples(streamer, inputs=['X1'], outputs=['Y1', 'Y2'])  # generates tuples of (list of ndarray, list of ndarray)
```

pescador/zmq_stream.py
```diff
@@ -181,18 +180,19 @@ def generate(self, max_batches=None):
...
 
             # Yield from the queue as long as it's open
             while True:
-                yield zmq_recv_batch(socket)
+                yield zmq_recv_data(socket)
 
         except StopIteration:
             pass
 
+        # TODO: Fix bare except
```
bmcfee: no, this is an intentional bare except that gets reraised.  IIRC, this is here because of some weirdness with transportable exceptions.

pescador/maps.py
ejhumphrey: cool, this is helpful. I'm not super familiar with the `fit` interface, so I was trying to get as far as possible from the docs.

pescador/zmq_stream.py
ejhumphrey: gotcha, I'll nuke the comment ... I was just following my linter blind

pescador/maps.py
bmcfee: ahahhahahahhaha

Documentation?

pescador/zmq_stream.py
bmcfee: feel free to add a pylint disable comment there.

pescador/maps.py
ejhumphrey: now that I'm reading this on a computer, rather than my phone, I've discovered that I've written the behavior you desire ... although perhaps in not the most elegant way. 

The *keys* get upcast to a list, but singleton items are downcast to np.ndarrays; see L112, L116.

This was the first semi-clean way I saw toward the end of the night to handle both string / list inputs, and return np.ndarray / list outputs.

ejhumphrey:
Alrighty, I think this PR is all buttoned up; tests pass, and fixed the drop in coverage. Just need signoff from y'all (@bmcfee & @cjacoby)

cjacoby:
Apologies; I'm running around like crazy this week in preparation for flying to NYC tomorrow. If i don't get to it today (unlikely, honestly), I'll catch up on all this on the plane tomorrow. 

If that's blocking you and you want to push ahead, either just do it or poke me "offline" and I'll bump it up in my queue.

pescador/buffered.py
```diff
@@ -46,6 +48,12 @@ def __init__(self, streamer, buffer_size,
              completing teh last batch, will just return the number
              of samples currently in the buffer.
         """
+        warn('`BufferedStreamer` is deprecated in 1.1.0 '
+             'This functionality is superseded by the generator function '
+             '`pescador.buffer_data` in 2.0.0, which can be used with '
+             '`pescador.Streamer` to similar ends.'
+             'Use this idiom instead to maintain forwards compatibility.',
+             DeprecationWarning)
```
bmcfee: Are the examples in the documentation up to date with the latest jazz?

pescador/core.py
```diff
@@ -35,54 +41,54 @@ class Streamer(object):
...
+    ...     print(i)  # Displays 0, 1, 2, 3, 4, 0, 1, 2, ...
 
     An alternate interface for the same:
 
-    >>> for i in GS(cycle=True):
-    ...     print(i)
+    >>> for i in stream(cycle=True):
+    ...     print(i)  # Displays 0, 1, 2, 3, 4, 0, 1, 2, ...
+
+    NOTE: We don't use cycle because it's a closure and can't be serialized.
```
bmcfee: this note shouldn't be in the docstring

pescador/core.py
```diff
@@ -213,40 +225,49 @@ def tuples(self, *items, **kwargs):
...
         See Also
         --------
-        generate
+        iterate
         cycle
-        tuples
         '''
+        max_iter = rename_kw('max_batches', max_batches,
+                             'max_iter', max_iter,
+                             '1.1.0', '2.0.0')
```
bmcfee: Please be consistent about major/minor version numbers in deprecations.  I see `1.1` and `2.0` elsewhere.

pescador/core.py
```diff
@@ -204,7 +213,10 @@ def tuples(self, *items, **kwargs):
         generate
         cycle
         '''
-
+        warn('`Streamer.tuples()` is deprecated in 1.1.0 '
+             'This functionality is moved to `pescador.tuples` in 2.0.0. '
```
bmcfee: is there actually a `pescador.tuples`?  I only see `maps.keras_tuples`

pescador/maps.py
bmcfee: I think we also want an old-style tuples map that works like `Streamer.tuples` does/did:

```python
def tuples(gen, *args):
    for item in gen:
         yield (item[x] for x in args)
```
or something

tests/test_core.py
```diff
@@ -1,23 +1,53 @@
...
 
-import pescador
+import pescador.core
 import test_utils as T
 
 
-def test_streamer_list():
+def test_StreamActivator():
+    with pytest.raises(pescador.core.PescadorError):
+        pescador.core.StreamActivator(range)
```
bmcfee: this only tests the failure mode?  I guess that's okay.

tests/test_maps.py
```diff
@@ -0,0 +1,62 @@
...
+            for n in range(10)]
+
+    stream = pescador.maps.keras_tuples(data, inputs="foo", outputs="bar")
+    for n, (x, y) in enumerate(stream):
+        assert n == x == y * 2
+
+    stream = pescador.maps.keras_tuples(data, outputs="whiz")
+    for n, (x, y) in enumerate(stream):
+        assert n == y / 2.
+        assert x is None
```
bmcfee: Should this be` None` or `[]`?

pescador/buffered.py
ejhumphrey: no, probably not. I'll look into that now

pescador/core.py
ejhumphrey: ha, right. that was a note for me to migrate into documentation elsewhere

pescador/core.py
ejhumphrey: nope, that's stale

pescador/core.py
ejhumphrey: is 1.1 fine? or 1.1.0?

tests/test_maps.py
ejhumphrey: [keras docs](https://keras.io/models/model/#fit) (currently) say `None`

pescador/core.py
bmcfee: i'd go with 1.1; sub-minors should never be involved in deprecations.

examples/frameworks/keras_example.py
```diff
@@ -93,71 +117,99 @@ def build_model(input_shape):
...
-    epochs=epochs,
-    verbose=1,
-    validation_data=(X_test, Y_test))
-
-score = model.evaluate(X_test, Y_test, verbose=0)
-print('Test score:', score[0])
-print('Test accuracy:', score[1])
+try:
+    model.fit_generator(
+        pescador.tuples(batches, 'X', 'y'),
```
bmcfee: `pescador.tuples(batches.cycle(), 'X', 'y')` ?

ejhumphrey: batches is infinite anyways?

ejhumphrey:
alrighty, updated the examples, and added some slightly more meaningful benchmarking to the zmq_example, which burns cycles to prove that things are happening in the background.

LET'S DO THIS

examples/frameworks/keras_example.py
```diff
@@ -93,71 +118,101 @@ def build_model(input_shape):
...
-def data_generator(X, y):
-    '''A basic generator for the data.'''
+    Yields
+    ------
+    data : dict
+        Single image sample, like {X: np.ndarray, y: np.ndarray}
+    '''
     X = np.atleast_2d(X)
-    # y's are binary vectors, and should be of shape (1, 10) after this.
+    # y's are binary vectors, and should be of shape (10,) after this.
```
bmcfee: atleast_2d will make them shape (1,10)

examples/mux_files_example.py
```diff
@@ -61,14 +61,14 @@ def npz_generator(npz_path):
     """Generate data from an npz file."""
     npz_data = np.load(npz_path)
     X = npz_data['X']
-    # y's are binary vectors, and should be of shape (1, 10) after this.
+    # y's are binary vectors, and should be of shape (10,) after this.
```
bmcfee: same comment here, i think?

pescador/maps.py
bmcfee: Upon closer inspection, I think this still isn't correct.

I don't think `fit(inputs=x)` and `fit(inputs=[x])` do the same thing (I might be wrong on that though).  If that's the case, then upcast-downcast will not behave as expected.

pescador/mux.py
```diff
@@ -83,14 +90,48 @@ def __init__(self, seed_pool, k,
 
             If None, the random number generator is the RandomState instance
             used by np.random.
+
+        seed_pool : iterable of streamers
+            .. warning:: This parameter name was deprecated in pescador 1.1
+            Use the `streamers` parameter instead.
+            The `seed_pool` parameter will be removed in pescador 2.0.
+
+        lam : float > 0.0
```
cjacoby: :100: on converting `lam` to `rate` - I think that's sooo much more intuitive.

However, you missed including `rate` in the docs here. I suspect the `lam` above is supposed to be `rate`.

pescador/version.py
```diff
@@ -3,4 +3,4 @@
 """Version info"""
 
 short_version = '1.0'
-version = '1.0.0'
+version = '1.1.0'
```
cjacoby: @bmcfee - Correct me if I misunderstand the case, but I believe last time I tried to do this that you requested all version increments be in a separate PR. I would expect that to apply here.

Also... shouldn't we increment the `short_version` too?

tests/test_buffered.py
```diff
@@ -0,0 +1,60 @@
...
+
+    gen_stream = pescador.Streamer(T.md_generator, dimension, 50,
+                                   size=batch_size)
+    estimate = pescador.BufferedStreamer(gen_stream, buf_size)
+
+    estimate = list(__unpack_stream(estimate))
+
+    T.__eq_lists(reference, estimate)
+
+    estimate = pescador.BufferedStreamer(gen_stream, buf_size)
```
cjacoby: curious; is this copypasta, or have you kept the old interface here on purpose? Same below.

(If these tests are here to make sure the old interface still works, that's fine, just mention that in a comment someplace, and that they should be removed when the deprecations are removed.)

examples/frameworks/keras_example.py
ejhumphrey: done

examples/mux_files_example.py
ejhumphrey: done

pescador/mux.py
ejhumphrey: yup, caught it

pescador/buffered.py
```diff
@@ -56,151 +63,30 @@ def activate(self):
...
         ----------
-        max_batches : int
-            For the BufferedStreamer, max_batches is the
+        max_iter : int
+            For the BufferedStreamer, max_iter is the
             number of *buffered* batches that are generated,
             not the number of individual samples.
+
+        partial : bool, default=True
+            Return buffers smaller than the requested size.
```
ejhumphrey: done

pescador/version.py
ejhumphrey: can I trouble y'all for consensus? I'd love to get this over the goal line today 😄

tests/test_buffered.py
ejhumphrey: ya, the whole `buffered` submodule is going away in 2.0, I've added a comment.

ejhumphrey:
couple comments here because this thread is getting *insane*.
- versioning: what's our stance (this vs separate)?
- docs: help would be great. also I dunno what I'm doing. so help would be *really* great.

pescador/version.py
cjacoby: Plz see: https://github.com/pescadores/pescador/blob/master/CONTRIBUTING.md, "Additional Comments".

cjacoby:
Re: Versioning. I'm posting this here, too 'cause embedded is getting too crazy.

The `CONTRIBUTING.md` says:
```Please keep branches as simple as possible: do not implement multiple features in the same branch, unless they are mutually dependent on one another. By the same token, please do not change unrelated portions of the repository (e.g., version numbers or changelog entries).```
and by that, I'd say, separate PR for version update.

Re: Docs - let's start an Issue with a checklist?

**Comment by cjacoby** [comment]
(Once we have those two things, I think this PR is 👍

File: pescador/version.py (hunk 21 above)
Comment (ejhumphrey) [inline pescador/version.py]: roger roger

**Comment by ejhumphrey** [comment]
hey team, two things: 
 - for clarity, you're saying create a docs issue **and** fix it in this PR? or create an issue and fix it elsewhere?
- in parallel, I started updating a recent project to (what will be) pescador 1.1, and having some difficulties dealing with `Mux`, specifically, the `Mux-of-Muxes` case that remains a little bit suspicious, as far as I'm concerned. I'm still in the process of determining whether my current struggles are a result of poor initial design of data streams in that project, a fundamental issue with the Mux API, or some combination of both. Regardless, I don't want to merge this PR until I've got concrete answers about what the hell is going on.

**Comment by cjacoby** [comment]
* Re: docs. I'm proposing it gets fixed in a separate PR. The items could get logged in that PR, or in an issue. But I'd start with an issue at least. If you write up an issue with all the components that changed such that they will affect the docs, I could take a stab at updating the docs to reflect the changes made in a PR. 

* Re: Muxes. This sounds like a good candidate for identifying via improved testing, ya?

File: pescador/mux.py [hunk 24]
```diff
@@ -19,59 +20,65 @@ class Mux(core.Streamer):
...
-            If ``True``, Streamers that previously produced no data are never
+        prune_empty_streams : bool
+            Disable streamers that produce no data.
+            If ``True``, streamers that previously produced no data are never
             revisited.
             Note that this may be undesireable for streams where past emptiness
             may not imply future emptiness.
 
         revive: bool
             If ``with_replacement`` is ``False``, setting ``revive=True``
```
ejhumphrey: @bmcfee @cjacoby, pop quiz ... if `with_replacement=True`, what is the significance of `revive`?

cjacoby: Nothing.

ejhumphrey:
@mcfee, can I trouble you for a ✅ ?

pescador/mux.py
```diff
@@ -104,45 +145,48 @@ def __init__(self, seed_pool, k,
...
 
         for idx in range(self.k):
 
-            if not (self.seed_distribution > 0).any():
+            if not (self.distribution > 0).any():
                 break
 
+            # TODO: Stream uniqueness can't be guaranteed here.
+            # Other machinery necessary to impose permutation, this is
+            # implicitly `with_replacement`
```
bmcfee: really?  I thought we zero out the probability of active streams when `with_replacement=False`.

ejhumphrey: You're correct, and I've already removed the comment locally; this was my read at midnight last night, and fixed it this am with some coffee-inspired clarity.

ejhumphrey: btw I was totally wrong and just thought i did (uhderp)

ejhumphrey:
replaced by #88

Issue #84, opened 2017-04-18
On Muxes


ejhumphrey:
Recent adventures in pescador development have resulted in some observations / questions about Muxes, their (desired) behavior.

Firstly, the sampling nature of a Mux can be put in three modes, controlled through two binary parameters (resulting in one "indifferent" state):
- `with_replacement=False`, `revive=False`: Run every stream once to exhaustion.
- `with_replacement=False`, `revive=True`: Run every stream indefinitely; there is no duplication at the level of active streams, but source streams are reused.
- `with_replacement=True`, `revive=*`: Streams are sampled independently and indefinitely; a single stream may be active multiple times.

From discussions with @cjacoby, there are some other use cases in which one would want to multiplex streams, but can't be achieved with this interface. To kick off the discussion, here are some that I'd like:

**Permuted Mux:** Streams are activated as a permutation of the input ordering, e.g. for n_streams=3, the activation ordering could be `stream_idx=[0, 2, 1, 2, 0, 1, 1, 2, 0, ...]`. 

**RoundRobin Mux:** Streams are deterministically multiplexed in the order provided, e.g. `stream_idx=[0, 1, 2, 0, 1, 2, ...]`.

What's interesting is that Permuted and RoundRobin Mux are pretty similar, and could be achieved with a shuffle parameter. In particular, I'm curious what other kinds of common behavior could be factored, which might lead to something like:

```
class BaseMux(Streamer):
    ...

class ShuffledMux(BaseMux):
    ...

class SampledMux(BaseMux):
    ...
```

...where `SampledMux` is actually the (Stochastic)`Mux` of today.

Thoughts? Others?

cjacoby:
Thought: A Mux in the abstract:

A `Streamer` which consumes/contains a list of streamers, and produces one sample from one of them at every step.

Components/methods of every streamer (which could be unit tested!):
 * a method for choosing the “active” streams.
 * a method for choosing a sample from the active streams
 * Handling streams that die.
 * ?

cjacoby:
# Mux Use Cases
(which we need to cover)

* I want to sample from N streams equally (Permuted/Shuffled Mux?)

* I want to sample from N streams, weighting how often we sample each stream differently.
    - Handle class imbalance, etc.
    - A `SampledMux` with uniqueness guarantees / constraints on how sampled.

* I have a very very large dataset, and I want to be able to pull subsets from it which I can fit in memory, and sub-sample from those, keeping only one or a small number open at a time.
[...]

PR #89, opened 2017-04-18
Fixes #86 -- Re-iterating Muxes

### LINKED ISSUE #86: Unable to re-iterate Muxes
After a day of struggles (and a night of bad pescadreams), I do believe I've identified the source of one of the issues I've been having.

**tl;dr:** When a mux expires *or* is interrupted, it can't return a fresh iterator for subsequent iteration.

This behavior is captured with [this test](https://github.com/pescadores/pescador/blob/ejh_20170403_test_refactor/tests/test_mux.py#L252), which produces the following mismatched output:
```
['a', 'd', 'e...'c', 'f', ...] == []
```
Here, the mux works once (left side), but can't return a fresh iterator to reproduce its sequence.

As for *why* this is happening ... not 100% sure yet, but I think it's because there's some iterator-dependent state getting set in `__init__` instead of `activate`, such as the [source stream distribution](https://github.com/pescadores/pescador/blob/ejh_20170403_test_refactor/pescador/mux.py#L152). May be other things, but won't be clear until digging in.

@bmcfee any thoughts on this? @cjacoby and I discussed, and the plan is to fix on a separate PR after #80 is fully back.

ejhumphrey:
Moves `Mux.distribution` to `Mux.distribution_` as active stream state, which is a more appropriate place for the data structure (since it's mutated while the Mux is active).

Open RFC regarding any other tests that'd be worth adding.

### DIFF
--- a/pescador/mux.py
+++ b/pescador/mux.py
@@ -148,11 +148,8 @@ class Mux(core.Streamer):
         if not self.n_streams:
             raise PescadorError('Cannot mux an empty collection')
 
-        # Set up the sampling distribution over streams
-        self.distribution = 1. / self.n_streams * np.ones(self.n_streams)
-
         if self.weights is None:
-            self.weights = self.distribution.copy()
+            self.weights = 1. / self.n_streams * np.ones(self.n_streams)
 
         self.weights = np.atleast_1d(self.weights)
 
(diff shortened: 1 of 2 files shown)


cjacoby:
LGTM. I'd be okay with tests improvement waiting till #84 work.

PR #97, opened 2017-06-13
Fixes #87

### LINKED ISSUE #87: Muxes can get caught in infinite loop.
In updating another project from pescador==0.1.3 to 1.0, I was running into an issue where a data stream was hanging indefinitely. 

Fortunately, having identified #86 (attempts to re-iterate a mux returned no samples), I was able to figure out why a mux-of-muxes would get caught in an infinite loop. 

**tl:dr;** If a Mux finds itself in a situation where all of its streams are fundamentally exhausted (cannot produce an iterator that yields samples), attempts to iterate will run forever. 

This is documented in a [test case here](https://github.com/pescadores/pescador/blob/ejh_20170403_test_refactor/tests/test_mux.py#L307). 

I *think* [a check here](https://github.com/pescadores/pescador/blob/ejh_20170403_test_refactor/pescador/mux.py#L209) that `distribution.any() > 0` would do the trick... @bmcfee any thoughts?

ejhumphrey:
Fix to avoid infinite loops on empty streams, issue #87

### DIFF
--- a/pescador/mux.py
+++ b/pescador/mux.py
@@ -71,8 +71,11 @@ class Mux(core.Streamer):
             Disable streamers that produce no data.
             If ``True``, streamers that previously produced no data are never
             revisited.
-            Note that this may be undesireable for streams where past emptiness
+            Note:
+            1. This may be undesireable for streams where past emptiness
             may not imply future emptiness.
+            2. Failure to prune truly empty streams with `revive=True` can
+            result in infinite looping behavior. Disable with caution.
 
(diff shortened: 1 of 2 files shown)


ejhumphrey:
hardest fought 13 lines of code in some time, ugh

ejhumphrey:
thx for the looksee!

agreed on the state-based testing ... i'll add an issue, and then we can decide if it's 1.1 milestone territory

Issue #49, opened 2016-12-28
Profile and speed up


cjacoby:
I did a bunch of experimenting about a month ago, and noticed that the Mux was significantly slowing down my sampling/training process.

Todo:
* Benchmark & analyze where the time is getting lost
* Fix it, if possible. At least understand it.

bmcfee:
Is your profiling code available somewhere?

cjacoby:
It was in a work thing; I'll try to refactor it into something sharable in
the next couple days. (I'm on the road to Seattle all day today, won't have
internet for real till tomorrow)

bmcfee:
Ping!  We should sync up on this some time.  It would be great to put out 1.0 in the near future.

cjacoby:
I agree. I've been thinking about this.

Want to hang about it soon? Free almost all of tomorrow and PT morning
Saturday.

cjacoby:
Thinking about making another `pescadores` repo to do this. `pescador-benchmarks` or something. Thoughts?

bmcfee:
> I did a bunch of experimenting about a month ago, and noticed that the Mux was significantly slowing down my sampling/training process.

Now that we have pytest in place, I threw pytest-profiling at it and plotted the results in snakeviz.  It looks like the biggest chunk is actually in random index sampling `np.random.choice`:

![image](https://cloud.githubusercontent.com/assets/1190540/23557446/1ac38c04-fffe-11e6-914c-854ecb1fcdf7.png)

note that the per-time cost is miniscule, but we call it 28604 times, so it adds up.

Otherwise, it looks like time is dominated by the underlying streamers/generators, which is what we'd expect.

bmcfee:
@ejhumphrey and I just had a chat, and it looks like the rng bottleneck could be easily accelerated by adding a buffering mechanism on top of rng.  I'll bench it and report back.

bmcfee:
The savings here are pretty minimal -- I think it may not be worth implementing.

ejhumphrey:
😞

bmcfee:
Random thought: @cjacoby are you sure it was `mux` and not `buffer` that was slowing down?  AFAICT, the mux overhead is relatively minimal, notwithstanding calls out to rng, and depending on your seed pool size, a bunch of vector math.  Buffering will typically invoke copies though, and that could for sure be a performance hit.

cjacoby:
I think you're probably correct; what you describe jives with my experience.

[...]

PR #98, opened 2017-06-14
Benchmarking & Regression test for buffer_stream & Streamer

cjacoby:
Implements changes requested in #49. More or less completed, with the following things still necessary before merge:

- [ ] choose actual parameters we will use for the test running (n_samples, sample_shapes, buffer_sizes)
- [ ] identify better thresholds, or better strategy for comparing changes in times between runs. It's pretty fragile right now.
- [ ] Add test vectors into pull request.
- [ ] Add test running into travis / `setup.py`

@ejhumphrey I'd appreciate it if you could weigh in on these points, or on anything further in this PR. Signing off for now.

### DIFF
--- a/tests/benchmark/benchmark_streamers.py
+++ b/tests/benchmark/benchmark_streamers.py
@@ -1,0 +1,199 @@
+"""benchmark_streamers is a test utility to generate benchmarks for
+different streamer configurations. It compares a standard generator vs a plain
+Streamer, and buffer_stream'd streamers of various buffer sizes.
+
+This utility is primarily a regression test, with the additional functionality
+of demonstrating potential weak points in the Pescador batching.
+"""
+import click
+import colorama
+import logging
+# import matplotlib.pyplot as plt
+import numpy as np
+import os
+import pandas as pd
+import pescador
+import tempfile
+import time
+
+logger = logging.getLogger("benchmark_streamers")
+
+
+sample_shapes = [(1, 10), (1, 10, 79)]
+n_samples = [100, 1000, 5000, 10000]
+buffer_sizes = [8, 32, 128, 256]
+# n_samples = [100, 1000, 5000, 10000, 20000, 25000, 50000]
+# sample_shapes = [(1, 10), (1, 10, 79), (1, 3, 28, 28), (1, 6, 26, 26)]
+# buffer_sizes = [8, 32, 128, 256]
+
+n_to_test = 1000
+regression_vector_path = os.path.join(os.path.dirname(__file__),
+                                      "benchmark_vectors.json")
+
+
+def statcode_colored(status_code, text=None):
+    if status_code == 0:
+        color = colorama.Fore.GREEN
+        print_text = "PASS"
+    else:
+        color = colorama.Fore.RED
+        print_text = "FAIL"
+
+    if text:
+        print_text = str(text) + ": " + print_text
+
+    return "{}{}{}".format(
+        color + colorama.Style.BRIGHT,
+        print_text,
+        colorama.Style.RESET_ALL)
+
+
+def create_npy(n_samples, sample_shapes, filename):
+    shape = list(sample_shapes)
+    shape[0] = n_samples
+    arrays_to_save = np.random.random(shape)
+
+    np.save(filename, arrays_to_save)
+
+
+def npy_gen(filename, mmap='c'):
+    data_in = np.load(filename, mmap_mode=mmap)
+
+    while True:
+        i = np.random.randint(len(data_in))
+        yield dict(X=np.array(data_in[i]))
+
+
+def benchmark_one(n_samples, sample_shape):
+    """Time sampling from a .npy with the passed configuration."""
+    key_str = "{}_{}".format(n_samples,
+                             "-".join([str(x) for x in sample_shape]))
+    filename = "random_{}.npy".format(key_str)
+
+    results = pd.Series(name=key_str)
+    with tempfile.TemporaryDirectory() as tmpdir:
+        filepath = os.path.join(tmpdir, filename)
+        # Create the test file
+        c0 = time.time()
+        create_npy(n_samples, sample_shape, filepath)
+        results['create_time'] = time.time() - c0
+        logger.info("{} created in {}s".format(filename,
+                                               results['create_time']))
+
+        # Sample from it with no streamer
+        sample_times = []
+        s0 = time.time()
+        for i, batch in enumerate(npy_gen(filepath)):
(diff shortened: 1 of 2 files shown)


tests/benchmark/benchmark_streamers.py
```diff
@@ -0,0 +1,199 @@
...
+different streamer configurations. It compares a standard generator vs a plain
+Streamer, and buffer_stream'd streamers of various buffer sizes.
+
+This utility is primarily a regression test, with the additional functionality
+of demonstrating potential weak points in the Pescador batching.
+"""
+import click
+import colorama
+import logging
+# import matplotlib.pyplot as plt
```
ejhumphrey: cruft?

tests/benchmark/benchmark_streamers.py
```diff
@@ -0,0 +1,199 @@
...
+import tempfile
+import time
+
+logger = logging.getLogger("benchmark_streamers")
+
+
+sample_shapes = [(1, 10), (1, 10, 79)]
+n_samples = [100, 1000, 5000, 10000]
+buffer_sizes = [8, 32, 128, 256]
+# n_samples = [100, 1000, 5000, 10000, 20000, 25000, 50000]
```
ejhumphrey: also cruft? (but part of your questions)

tests/benchmark/benchmark_streamers.py
```diff
@@ -0,0 +1,199 @@
+"""benchmark_streamers is a test utility to generate benchmarks for
+different streamer configurations. It compares a standard generator vs a plain
+Streamer, and buffer_stream'd streamers of various buffer sizes.
+
+This utility is primarily a regression test, with the additional functionality
+of demonstrating potential weak points in the Pescador batching.
+"""
+import click
```
ejhumphrey: interesting ... wasn't familiar with this. much preferred to argparse? or just for simple things?

also, standard lib now? or are you picking this up from conda?

tests/benchmark/benchmark_streamers.py
```diff
@@ -0,0 +1,199 @@
...
+
+def create_npy(n_samples, sample_shapes, filename):
+    shape = list(sample_shapes)
+    shape[0] = n_samples
+    arrays_to_save = np.random.random(shape)
+
+    np.save(filename, arrays_to_save)
+
+
+def npy_gen(filename, mmap='c'):
```
ejhumphrey: thoughts on also changing the shape / size of data read from memory? random slices and whatnot? this is where some of the slowdown comes from in my experience

tests/benchmark/benchmark_streamers.py
```diff
@@ -0,0 +1,199 @@
...
+        time_difference = (results_df - old_vectors).drop([
+            "create_time", "generator"])
+
+        time_diff_absolute = time_difference.loc[
+            time_difference.index.map(lambda x: 'per sample' not in x)]
+        time_diff_per_sample = time_difference.loc[
+            time_difference.index.map(lambda x: 'per sample' in x)]
+
+        # Check that the absolute times haven't increased by more than 10%
+        #  TODO: 10%? what's a good number here.
```
ejhumphrey: I worry somewhat about how this will translate across machines (or potentially VMs), where disk I/O will be variable? is that a founded concern?

ejhumphrey:
overall, this looks great. 

One idea that might help provide clarity around the questions raised would be to enumerate the specific behaviors that we're interested in benchmarking. For me, the big one is multi-file throughput, which is really stress-testing the Mux.

Would we want to add that here? or in a similar benchmarking test?

in other news ... I currently have no good ideas about resolving fragility of the expected outcomes. lemme mull that over.

tests/benchmark/benchmark_streamers.py
cjacoby: Well, maybe. I had this notion that I might give it the option to dump plots to disk, and if I do that, it's not cruft. But, that's certainly not a high priority.

tests/benchmark/benchmark_streamers.py
cjacoby: Yeah; I think the big ones are useful for identifying some behaviors, but also take a long time. (So I'm using the small ones for testing / proof of concept)

tests/benchmark/benchmark_streamers.py
cjacoby: Sure, but I don't necessarily think that's a Pescador related thing. But I could see the numbers being useful, certainly.

tests/benchmark/benchmark_streamers.py
cjacoby: Oh I think it certainly is a founded concern. I just have no ideas yet on how to deal with it. Maybe worth taking a look at pytest-benchmark's code? I dunno.

cjacoby:
I agree that it would be nice to do the same thing with `Mux`, but I think it should be a separate from testing the `buffer_stream`. I'm focusing on testing `buffer_stream`, because in my experience it's the thing that seemed to be slowing things down, and I was trying to prove that. (Whereas Mux seemed to be just fine). I'd rather get a good understanding that the buffering is solid before jumping into mux (although if you wanted to do mux in parallel, that might be easier.)

tests/benchmark/benchmark_streamers.py
cjacoby: I've been using it a bit lately. Found it on [awesome python](https://github.com/vinta/awesome-python), and I'm digging it a lot for simple use cases. 

I added it to the `setup.py` under `test_deps`, which I think should do the trick.

ejhumphrey:
related: **tl;dr** from a convo with @bmcfee – benchmarking and profiling routines should be included as tools in the repository, but not incorporated in the CI test suite. This wasn't necessarily obvious to me, but this is the sane answer.

cjacoby:
Re: tools in the repo. Does this therefore mean they should really be in a separate repo in the org?

ejhumphrey:
oooh, interesting proposal. I like that approach.

always forget that this has an org now...

cjacoby:
Okay, then I'll move this into a separate repo in the org (I think I started it already a while ago).

But that still leaves the matters of:
- [ ] choose actual parameters we will use for the test running (n_samples, sample_shapes, buffer_sizes)
- [ ] identify better thresholds, or better strategy for comparing changes in times between runs. It's pretty fragile right now.

Which I don't have a good answer for yet.

bmcfee:
If the consensus is to move this pr out to a separate repo in the org, should we close this pr?

cjacoby:
👍 Close this and move to separate repo.

Issue #81, opened 2017-04-12
Guaranteeing Mux statistics / behavior


ejhumphrey:
**tl;dr:** I think our `Mux` tests need to be more rigorous to make sure we're getting back the statistics one would expect from the parameters used to initialize the object.

I was looking into #79, and [decided to test](https://github.com/pescadores/pescador/blob/ejh_20170403_test_refactor/tests/test_mux.py#L153) that this issue was resolved in the midst of a broader refactor. While things mostly worked, I observed some behavior I didn't immediately understand. Basically, if I ended up with a bad RNG state, I'd end up with two duplicate streams of one sub-mux, and nothing from the other.

Okay, user error -- increasing the oversampling (``k=2 -> 10, max_iter=10->100``) "fixes" it, but this is a dumb hack and so I tried to do the correct thing, i.e. follow the docstring:

> If ``with_replacement`` is ``False``, setting ``revive=True``will re-insert previously exhausted streams into the candidate set.  This configuration allows a stream to be active at most once at any time.

However, this *also* didn't give me the behavior I wanted (at least, [not always](https://travis-ci.org/pescadores/pescador/jobs/221198293#L991)). Checking the tests, it seems that ``test_mux_revive`` doesn't inspect the stream it gets back, and I'm suspicious that Mux might not *always* be doing the thing we expect.

I'm going to explore with more testing, but in parallel -- is this perhaps surprising or consistent with others' observations? @bmcfee, @cjacoby?

ejhumphrey:
kay, few quick pre-bedtime findings:

1. Mux-of-Mux issues seem to be an issue of repeat iteration. 

For example, this causes problems:
```python
abc = pescador.Streamer('abc')
xyz = pescador.Streamer('xyz')
mux1 = pescador.Mux([abc, xyz], k=2, rate=None, revive=True,
                    with_replacement=False, prune_empty_streams=False)
samples1 = mux1.iterate(max_iter=1000)  # Fine

n123 = pescador.Streamer('123')
n456 = pescador.Streamer('456')
mux2 = pescador.Mux([n123, n456], k=2, rate=None, revive=True,
                    with_replacement=False, prune_empty_streams=False)
samples2 = mux2.iterate(max_iter=1000)  # Fine

mux3 = pescador.Mux([mux1, mux2], k=2, rate=None,
                    with_replacement=False, revive=True,
                    prune_empty_streams=False)
samples3 = mux3.iterate(max_iter=1000)  # ...hangs?
```

[...]

PR #100, opened 2017-07-06
Tests for Mux statistics

ejhumphrey:
Adds a test that stacked bootstrapped streams are approximately uniform at large N.

Note: looking over Mux tests, we've accumulated pretty decent coverage the the sampling statistics of the Mux are behaving as advertised. I think some of the sanding and polish that's been thrown at pescador in the last several weeks has really ironed out some of the subtle kinks that were empirically felt / observed. This PR is more symbolic as a seal of approval that we're confident in our work.

### DIFF
--- a/tests/test_mux.py
+++ b/tests/test_mux.py
@@ -161,7 +161,6 @@ def test_mux_of_muxes_itered():
                         random_state=135)
     samples1 = mux1.iterate(max_iter=1000)
     count1 = collections.Counter(samples1)
-    print(count1)
     assert set('abcxyz') == set(count1.keys())
 
     n123 = pescador.Streamer('123')
@@ -171,7 +170,6 @@ def test_mux_of_muxes_itered():
                         random_state=246)
     samples2 = mux2.iterate(max_iter=1000)
     count2 = collections.Counter(samples2)
-    print(count2)
     assert set('123456') == set(count2.keys())
 
     # Note that (random_state=987, k=2) fails.
@@ -180,7 +178,6 @@ def test_mux_of_muxes_itered():
                         random_state=987)
     samples3 = mux3.iterate(max_iter=1000)
     count3 = collections.Counter(samples3)
-    print(count3)
     assert set('abcxyz123456') == set(count3.keys())
 
 
@@ -203,19 +200,20 @@ def test_mux_of_muxes_single():
                         prune_empty_streams=False)
     samples3 = list(mux3.iterate(max_iter=10000))
     count3 = collections.Counter(samples3)
-    print(samples3[:10], count3)
     assert set('abcxyz123456') == set(count3.keys())
 
 
 def test_critical_mux():
     # Check on Issue #80
     chars = 'abcde'
-    streamers = [pescador.Streamer(x * 5) for x in chars]
+    n_reps = 5
+    streamers = [pescador.Streamer(x * n_reps) for x in chars]
     mux = pescador.Mux(streamers, k=len(chars), rate=None,
-                       with_replacement=False, revive=True,
+                       with_replacement=False, revive=False,
                        prune_empty_streams=False, random_state=135)
-    samples = mux.iterate(max_iter=1000)
-    print(collections.Counter(samples))
+    samples = list(mux.iterate(max_iter=1000))
+    assert len(collections.Counter(samples)) == len(chars)
+    assert len(samples) == len(chars) * n_reps
 
 
 def _choice(vals):
@@ -251,7 +249,6 @@ def test_critical_mux_of_rate_limited_muxes():
     count = collections.Counter(samples)
     max_count, min_count = max(count.values()), min(count.values())
     assert (max_count - min_count) / max_count < 0.2
-    print(count)
     assert set('abcdefghijkl') == set(count.keys())
 
 
@@ -275,7 +272,7 @@ def test_sampled_mux_of_muxes():
     # And inspect the first mux
     samples1 = list(mux1(max_iter=6 * 10))
     count1 = collections.Counter(samples1)
-    print(count1)
+
     assert set(count1.keys()) == set('abcdef')
 
     # Build another set of streams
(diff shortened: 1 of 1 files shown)


ejhumphrey:
addresses #81

tests/test_mux.py
```diff
@@ -310,3 +312,35 @@ def test_mux_inf_loop():
...
+
+    flat_mux = pescador.Mux([ab, cd, ef, gh, ij, kl], k=6, rate=None,
+                            with_replacement=False, revive=False)
+
+    max_iter = 50000
+    samples1 = list(stacked_mux.iterate(max_iter=max_iter))
+    samples2 = list(flat_mux.iterate(max_iter=max_iter))
+    count1 = collections.Counter(samples1)
+    count2 = collections.Counter(samples2)
+    print(count1, count2)
```
bmcfee: remove this print

tests/test_mux.py
```diff
@@ -310,3 +312,35 @@ def test_mux_inf_loop():
...
     assert len(list(mux(max_iter=100))) == 0
+
+
+def test_mux_stacked_uniform_convergence():
+    ab = pescador.Streamer(_choice, 'ab')
+    cd = pescador.Streamer(_choice, 'cd')
+    ef = pescador.Streamer(_choice, 'ef')
+    mux1 = pescador.Mux([ab, cd, ef], k=2, rate=2,
+                        with_replacement=False, revive=True)
+
```
bmcfee: these muxes should be seeded to avoid random test failure.

bmcfee:
Looking good so far, thanks for this!  I think we can make it more rigorous without much effort though (see review comments).

tests/test_mux.py
cjacoby: 👍

tests/test_mux.py
```diff
@@ -310,3 +312,35 @@ def test_mux_inf_loop():
                        with_replacement=False, random_state=1234)
 
     assert len(list(mux(max_iter=100))) == 0
+
+
+def test_mux_stacked_uniform_convergence():
```
cjacoby: A 1-sentence summary docstring here might be nice. I like these for more complex tests (like this one).

tests/test_mux.py
cjacoby: Or, make it a `logger.debug()`/`logger.info()`. But definitely not print

tests/test_mux.py
ejhumphrey: done

tests/test_mux.py
ejhumphrey: done

ejhumphrey:
i can haz final signoff? plznthx

tests/test_mux.py
```diff
@@ -310,3 +312,35 @@ def test_mux_inf_loop():
...
+    max_iter = 50000
+    samples1 = list(stacked_mux.iterate(max_iter=max_iter))
+    samples2 = list(flat_mux.iterate(max_iter=max_iter))
+    count1 = collections.Counter(samples1)
+    count2 = collections.Counter(samples2)
+    print(count1, count2)
+    assert set('abcdefghijkl') == set(count1.keys()) == set(count2.keys())
+    c1, c2 = [list(c.values()) for c in (count1, count2)]
+    np.testing.assert_almost_equal(
+        np.std(c1) / max_iter, np.std(c2) / max_iter, decimal=2)
```
ejhumphrey: 1. the two distributions should have similar statistics, i.e. approx equal means and variances.
2. order of keys shouldn't matter; the `N` characters should occur with similar frequency, and I'm just checking the distributions.
3. normalizing out `max_iter` ... otherwise it grows with count, no?
4. I happened to fix (or at least change) this since you've commented. Now it's a single uniformly sampled stream. Lemme know if you still take issue with it.

ejhumphrey:
looks like we hit a race condition on edits versus comments...

tests/test_mux.py
bmcfee: In reverse order:

4. Yeah, I think I still take issue (sorry! :grimacing:)  The problem I have is that you're comparing two potentially unknown quantities (flat and nested muxes) to each-other, rather than comparing each one to a known reference with the desired properties.

3. `std(X)/N` just seems like a weird quantity for count data.  I think a better test would be to normalize the count distribution first, so you have a vector of observation frequencies, and then compare it to the uniform distribution, eg by chi^2 or KL-divergence or whatever.  Aggregating all counts before comparison seems wrong to me.
2. See above: test should be coordinate-wise, not in aggregate.
1. mean and std are fine for gaussian data, but not generally sufficient for arbitrary distributions

ejhumphrey: kay, will fix

tests/test_mux.py
```diff
@@ -330,17 +330,17 @@ def test_mux_stacked_uniform_convergence():
...
-    count1 = collections.Counter(samples1)
-    count2 = collections.Counter(samples2)
+    chars = 'abcdefghijkl'
+    samples = list(stacked_mux.iterate(max_iter=max_iter))
+    counter = collections.Counter(samples)
+    assert set(chars) == set(counter.keys())
+
+    counts = np.array(counter.values())
+    exp_count = float(max_iter / len(chars))
+    max_error = np.max(np.abs(counts - exp_count) / exp_count)
```
bmcfee: I think this would be cleaner using `scipy.stats`:
```python
counts = np.array(counter.values())
assert scipy.stats.chisquare(counts).pvalue >= 0.99
```
(default for `chisquare` is to compare against a uniform distribution)

ejhumphrey:
thx for the protip, pub time in the UK now though...

bmcfee:
I took the liberty of fixing this build error; should be good to merge?  I'm not gonna be a stickler for the statistical tests.

bmcfee:
I noticed some stochastic failures on this one, since the `_choice` helper function was not seeded.  I fixed the seed, but then noticed that it was failing for the 0.5% error test, so I went back and replaced it with a chi^2 test at p-value >= 0.95.

I think this is all now in order, and ready to :shipit:

ejhumphrey:
👍  lgtm

PR #103, opened 2017-10-09
Mux Refactor

cjacoby:
This is a work in progress, but far enough where comments would be useful. Docstrings may or may not be accurate / up to date; I'll work on that next. Warning: there's kind of a lot going on here. Will try to summarize more completely in the "real" PR once the following is done.

Still need to:
- [x] Mux Docstrings & docs
    - [x] Update / rewrite
    - [x] Proofread
- [x] Expand tests to cover new modes.

@bmcfee if you have ideas on tests (or want to take a stab at adding some coverage), I'd appreciate  anything.

### DIFF
--- a/pescador/mux.py
+++ b/pescador/mux.py
@@ -395,13 +395,8 @@ class BaseMux(core.Streamer):
 
     @property
     def n_streams(self):
-        """Return the number of streamers. Will fail if it's an iterable,
-        in which case just return None.
-        """
-        try:
-            return len(self.streamers)
-        except TypeError:
-            return None
+        """Return the number of streamers."""
+        return len(self.streamers)
 
     def activate(self):
         """Activates the mux as a streamer, choosing which substreams to
--- a/pescador/mux.py
+++ b/pescador/mux.py
@@ -1,5 +1,89 @@
 #!/usr/bin/env python
-'''Stream multiplexing'''
+'''Stream multiplexing
+
+(Refactoring thoughts; TODO - Remove)
+
+Goals:
+
+ 1. A `BaseMux` which defines the required interface for a Mux.
+
+    What is a Mux?
+
+        A `Mux` is at its simplest a Streamer which wraps N other streamers,
+        and at every step yields a sample from one of its sub-streamers.
+
+        In that sense, the BaseMux must have the following functions
+
+        `__init__(self, streamers)`
+        `iterate(self, max_iter)`
+        `activate()`
+        `deactivate()`
+        `_next_sample_index()` - chooses what stream to yield from
+
+ 2. BaseMux Subclasses
+
+    a. `PoissonMux`
+
+        Essentially identical to the previous `Mux`. `PoissonMux` will have
+        a single `mode` parameter which selects how it operates
+
+        `with_replacement` (with_replacement=True, revive=*,
+                            prune_empty_streams=False):
+            All streams are active, and revive if exhausted. Kind of the
+            opposite of `exhausted`
+
+        `single_active` (with_replacement=False, revive=True) :
+
+            This setting makes it so that once a stream from the pool
+            of streamers is activated, it can be used at most once.
+            Streams are revived when they are exhausted.
+
+            Propose to name this something like `unique_active`.
+
+        `exhaustive` (with_replacement=False, revive=False) :
+            Sample the streams until they all die, and then stop.
+
+            -> Does this need to raise StopIteration? We are not doing that now
+
+        * Questions:
+            - how does prune_empty_streams work in here? Currently, none
+             of the intermediate checks also check anything against
+             valid_streams_, making it potentially dangerous / ineffective in
+             (revive & ~with_replacement) mode.
+
+             I think this is not a problem in `with_replacement` mode,
+             just in `unique_active` mode.
+
+             Basically, there is a separate parameter here that seems
+             to control *actually* pruning empty seeds, which
+             seems like it should actually be:
+                * revive == True revive seeds
+                * revive == False prune empty seeds.
+
+            So, do we:
+             * turn with_replacement/revive+prune_empty_seeds into
+              one single mode parameter?
+
+    b. `ShuffledMux`
+
+        k == N; draws one sample from each stream, guaranteeing to
+        choose them with equal probability.
+
+        Could have some options for what to do if they die (revive / prune)
+
+    c. `RoundRobinMux`
+
+        AKA RoundRobin - Similar to `ShuffledMux` but guaranteed to
+        iterate over the streamers in strict order.
+
+    d. `ChainMux`
+
+        As in itertools.chain(). Runs the first streamer to exhaustion, then
+        the second, then the third, etc. k=1
+
+ 3. Old Mux still works but is deprecated
+
+'''
 import six
 import numpy as np
 
@@ -231,8 +315,12 @@ class Mux(core.Streamer):
                         self.valid_streams_[self.stream_idxs_[idx]] = False
 
                     if self.revive and not self.with_replacement:
-                        # If we need to revive a seed, give it the max
-                        # current probability
+                        # When this case is hit, the `distribution_` for
+                        # this "seed"/"stream" is 0.0, because it got set
+                        # to when we activated it. (in `_new_stream`)
+
+                        # Since revive mode is on, we set it to the max
+                        # current probability to enable it to be used again.
                         if self.distribution_.any():
                             self.distribution_[self.stream_idxs_[idx]] = (
                                 np.max(self.distribution_))
@@ -286,49 +374,17 @@ class Mux(core.Streamer):
             n_stream = None
 
         # If we're sampling without replacement, zero this one out
+        # This effectively disables this stream as soon as it is chosen,
+        # preventing it from being chosen again (unless it is revived)
         if not self.with_replacement:
             self.distribution_[idx] = 0.0
 
+            # Correct the distribution
             if (self.distribution_ > 0).any():
                 self.distribution_[:] /= np.sum(self.distribution_)
 
         return (self.streamers[idx].iterate(max_iter=n_stream),
                 self.weights[idx])
-
-
-class ShuffledMux(Mux):
-    """A variation on a mux, which takes N streamers, and samples
-    from them equally, guaranteeing all N streamers to be "active",
-    unlike the base Mux, which randomly chooses streams when activating.
-
-    TODO Better name / more general approach for this.
-    """
-    def __init__(self, streamers):
-        super(ShuffledMux, self).__init__(
-            streamers, k=len(streamers),
-            revive=True, rate=None)
-
-    def activate(self):
-        """Activates a number of streams"""
-        self.distribution_ = 1. / self.n_streams * np.ones(self.n_streams)
-        self.valid_streams_ = np.ones(self.n_streams, dtype=bool)
-
-        self.streams_ = [None] * self.k
-
-        self.stream_weights_ = np.zeros(self.k)
-        self.stream_counts_ = np.zeros(self.k, dtype=int)
-        # Array of pointers into `self.streamers`
-        self.stream_idxs_ = np.arange(self.k, dtype=int)
-
-        for idx in range(self.k):
-
-            if not (self.distribution_ > 0).any():
-                break
-
-            self.streams_[idx], self.stream_weights_[idx] = (
-                self._new_stream(self.stream_idxs_[idx]))
-
-        self.weight_norm_ = np.sum(self.stream_weights_)
 
 
 class BaseMux(core.Streamer):
@@ -340,12 +396,13 @@ class BaseMux(core.Streamer):
 
      * When "activated", choose a subset of available streamers to stream from
        (the "active substreams")
-     * When a sample is pulled from the mux, choose which active substream to
-       stream from.
+     * When a sample is drawn from the mux (via generate),
+       chooses which active substream to stream from.
      * Handles exhaustion of streams (restarting, replacing, ...)
 
     """
-    def __init__(self, streamers, weights=None, random_state=None):
+    def __init__(self, streamers, k, weights=None, random_state=None,
+                 prune_empty_streams=True):
         """
         Parameters
         ----------
@@ -370,28 +427,209 @@ class BaseMux(core.Streamer):
             If None, the random number generator is the RandomState instance
             used by np.random.
         """
-        pass
+        self.streamers = streamers
+        self.n_streams = len(streamers)
+        self.k = k
+        self.prune_empty_streams = prune_empty_streams
+        self.weights = weights
+
+        # Clear state and reset actiave/deactivate params.
+        self.deactivate()
+
+        # TODO: Remove from base
+        # These are defaults; set them in the subclass.
+        self.with_replacement = True
+        self.revive = False
+
+        if random_state is None:
+            self.rng = np.random
+        elif isinstance(random_state, int):
+            self.rng = np.random.RandomState(seed=random_state)
+        elif isinstance(random_state, np.random.RandomState):
+            self.rng = random_state
+        else:
+            raise PescadorError('Invalid random_state={}'.format(random_state))
+
+        if not self.n_streams:
+            raise PescadorError('Cannot mux an empty collection')
+
+        if self.weights is None:
+            self.weights = 1. / self.n_streams * np.ones(self.n_streams)
+        self.weights = np.atleast_1d(self.weights)
+
+        if len(self.weights) != len(self.streamers):
+            raise PescadorError('`weights` must be the same '
+                                'length as `streamers`')
+
+        if not (self.weights > 0.0).any():
+            raise PescadorError('`weights` must contain at least '
+                                'one positive value')
+
+        self.weights /= np.sum(self.weights)
 
     def activate(self):
         """Activates the mux as a streamer, choosing which substreams to
         select as active."""
-        pass
+        # These do not depend on the number of streams, k
+        # This function must be filled out in more detail in a child Mux.
+        self.distribution_ = 1. / self.n_streams * np.ones(self.n_streams)
+        self.valid_streams_ = np.ones(self.n_streams, dtype=bool)
+
+        # The active streamers
+        self.streams_ = [None] * self.k
+
+        # Weights of the active streams.
+        # Once a stream is exhausted, it is set to 0
+        self.stream_weights_ = np.zeros(self.k)
+        # How many samples have been draw from each (active) stream.
+        self.stream_counts_ = np.zeros(self.k, dtype=int)
+        # Array of pointers into `self.streamers`
+        self.stream_idxs_ = np.zeros(self.k, dtype=int)
+
+        # Initialize each active stream.
+        for idx in range(self.k):
+
+            if not (self.distribution_ > 0).any():
+                break
+
+            # Setup a new streamer at this index.
+            self._new_stream(idx)
+
+        self.weight_norm_ = np.sum(self.stream_weights_)
 
     def deactivate(self):
         """Reset the Mux state."""
-        pass
+        self.streams_ = None
+        self.stream_idxs_ = None
+        self.stream_counts_ = None
+        self.stream_weights_ = None
+        self.weight_norm_ = None
 
     def iterate(self, max_iter=None):
         """Yields items from the mux."""
+        self._validate_subclass()
+
+        if max_iter is None:
+            max_iter = np.inf
+
+        with core.StreamActivator(self):
+            # Main sampling loop
+            n = 0
+
+            while n < max_iter and self.weight_norm_ > 0.0:
+                # Pick a stream from the active set
+                idx = self._next_sample_index()
+
+                # Can we sample from it?
+                try:
+                    # Then yield the sample
+                    yield six.advance_iterator(self.streams_[idx])
+
+                    # Increment the sample counter
+                    n += 1
+                    self.stream_counts_[idx] += 1
+
+                except StopIteration:
+                    # Oops, this stream is exhausted.
+
+                    # If we're disabling empty seeds, see if this stream
+                    # produced data at any point; if it didn't, turn it off.
+                    #  (Note) prune_empty_streams applies to all Muxes?
+                    if (self.prune_empty_streams and
+                            self.stream_counts_[idx] == 0):
+                        self.distribution_[self.stream_idxs_[idx]] = 0.0
+                        self.valid_streams_[self.stream_idxs_[idx]] = False
+
+                    # Call child-class exhausted-stream behavior
+                    self._on_stream_exhausted(idx)
+
+                    # If there are active streams reamining,
+                    # choose a new one to make active.
+                    if (self.distribution_ > 0).any():
+                        # Replace it and move on if there are still seeds
+                        # in the pool.
+                        self.distribution_[:] /= np.sum(self.distribution_)
+
+                        # Setup a new streamer at this index.
+                        self._new_stream(idx)
+                    else:
+                        # Otherwise, this one's exhausted.
+                        # Set its probability to 0
+                        self.stream_weights_[idx] = 0.0
+
+                    self.weight_norm_ = np.sum(self.stream_weights_)
+
+                # If everything has been pruned, kill the while loop
+                if not self.valid_streams_.any():
+                    break
+
+    def _on_stream_exhausted(self, idx):
+        """Override this to provide a Mux with additional behavior
+        when a stream is exhausted. This gets called *after* streams
+        are pruned.
+
+        Parameters
+        ----------
+        idx : int, [0:k - 1]
+            Index of the exhausted stream (in `self.stream_idxs_`).
+        """
         pass
 
-    def _activate_substreams(self):
-        """Chooses substreams to activate."""
-        pass
+    def _activate_stream(self, idx):
+        '''Randomly select and create a stream.
 
-    def _next_stream(self):
-        """Returns the stream index for the next sample"""
-        pass
+        Parameters
+        ----------
+        idx : int, [0:n_streams - 1]
+            The stream index to replace
+        '''
+        if len(self.streamers) != len(self.weights):
+            raise PescadorError('`streamers` must have the same '
+                                'length as `weights`')
+
+        if len(self.streamers) != len(self.distribution_):
+            raise PescadorError('`streamers` must have the same '
+                                'length as `distribution`')
+
+        # instantiate
+        if self.rate is not None:
+            n_stream = 1 + self.rng.poisson(lam=self.rate)
+        else:
+            n_stream = None
+
+        ## TODO: migrate this to child class
+        # If we're sampling without replacement, zero this one out
+        # This effectively disables this stream as soon as it is chosen,
+        # preventing it from being chosen again (unless it is revived)
+        if not self.with_replacement:
+            self.distribution_[idx] = 0.0
+
+            # Correct the distribution
+            if (self.distribution_ > 0).any():
+                self.distribution_[:] /= np.sum(self.distribution_)
+
+        return (self.streamers[idx].iterate(max_iter=n_stream),
+                self.weights[idx])
+
+    def _new_stream_index(self, idx=None):
+        """Returns an index of a streamer from `self.streamers` which
+        will get added to the active set.
+
+        Parameters
+        ----------
+        idx : int or None
+            The index is passed along so a child class can use it.
+            (The index is not required for a random stream as in PoissonMux,
+             but would be required for RoundRobin mux).
+        """
+        raise NotImplementedError("_new_stream_index() must be implemented in"
+                                  " a child class.")
+
+    def _next_sample_index(self):
+        """Returns the index in self.streams_ for which to try to draw
+        the next sample."""
+        raise NotImplementedError("_next_sample_index() must be implemented in"
+                                  " a child class.")
 
     def _new_stream(self, idx):
         '''Randomly select and create a stream.
@@ -401,7 +639,13 @@ class BaseMux(core.Streamer):
         idx : int, [0:n_streams - 1]
             The stream index to replace
         '''
-        pass
+        self.stream_idxs_[idx] = self._new_stream_index(idx)
+
+        # => self._activate_stream(idx)
+        self.streams_[idx], self.stream_weights_[idx] = (
+            self._activate_stream(self.stream_idxs_[idx]))
+
+        self.stream_counts_[idx] = 0
 
 
 class PoissonMux(BaseMux):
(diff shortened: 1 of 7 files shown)


bmcfee:
@cjacoby happy to look it over, but I think it might go quicker if the docstrings are revised for correctness (if not completeness) before I dive in.  That way I'll have a better idea of what to expect.

cjacoby:
Roger that; I'll try to get to that tomorrow AM.

cjacoby:
Alright - that should do the trick.

pescador/mux.py
```diff
@@ -1,5 +1,57 @@
...
+
+`ChainMux`
+
+    As in itertools.chain(), runs the first streamer to exhaustion, then
+    the second, then the third, etc. Uses only a single stream at a time.
+
+`Mux`
+
+    The pescador<2.0 `Mux` is still available and works the same,
+    but is deprecated.
```
bmcfee: We should have a deprecation decorator on `Mux`.  Maybe using https://github.com/amueller/futurepast ?

pescador/mux.py
```diff
@@ -263,7 +319,7 @@ def iterate(self, max_iter=None):
                 if not self.valid_streams_.any():
                     break
 
-    def __new_stream(self, idx):
+    def _new_stream(self, idx):
```
bmcfee: Is there any reason to move from `__` to `_` on these guys if we're not inheriting from `BaseMux`?

pescador/mux.py
```diff
@@ -286,11 +342,575 @@ def __new_stream(self, idx):
...
+
+     * When "activated", choose a subset of available streamers to stream from
+       (the "active substreams")
+     * When a sample is drawn from the mux (via generate),
+       chooses which active substream to stream from.
+     * Handles exhaustion of streams (restarting, replacing, ...)
+
+    """
+    def __init__(self, streamers, k, weights=None, random_state=None,
+                 prune_empty_streams=True):
```
bmcfee: Not all streamers have all of these parameters.  Should this not be:
```python
def __init__(self, streamers, random_state=None):
    ...
```
and then let the derived classes change the init signature?

pescador/mux.py
```diff
@@ -286,11 +342,575 @@ def __new_stream(self, idx):
...
+            self.rng = random_state
+        else:
+            raise PescadorError('Invalid random_state={}'.format(random_state))
+
+        if not self.n_streams:
+            raise PescadorError('Cannot mux an empty collection')
+
+        if self.weights is None:
+            self.weights = 1. / self.n_streams * np.ones(self.n_streams)
+        self.weights = np.atleast_1d(self.weights)
```
bmcfee: not sure `atleast_1d` is what we want here; it should be exactly 1D

bmcfee: Thinking more on this: should `BaseMux` be in the business of setting weights at all?  Since not all muxen need it, I'd say that should be punted upstream.

pescador/mux.py
```diff
@@ -286,11 +342,575 @@ def __new_stream(self, idx):
...
+        for idx in range(self.k):
+
+            if not (self.distribution_ > 0).any():
+                break
+
+            # Setup a new streamer at this index.
+            self._new_stream(idx)
+
+        self.weight_norm_ = np.sum(self.stream_weights_)
+
```
bmcfee: I think I don't 100% understand why `activate` is necessary in general, and as a public method.

It seems like the semantics of streamer are such that `activate/deactivate` happen at the beginning (and end?) of calls to `iterate`, and there's no reason to externally reset a streamer.

I bring this up because if we can get away with eliminating these methods, we could potentially isolate generator state to the iterator method, and make it possible to have the mux support multiple simultaneous iteration.

This isn't a requirement, but I wanted to question this while we're in the process of redoing the API.

pescador/mux.py
cjacoby: I have done this; question though; how do i include / import `futurepast` using `setup.py`, since it's not in `pip`? I know how to do it with `pip install -r requirements.txt`, but not `setup.py`.

pescador/mux.py
cjacoby: ... I do not remember why I did this, or if i had any reasoning for doing so. 

A la [1], I think you are right, and that, especially the way that I have it set up now, `_new_stream(self, idx)` is *always* the `BaseMux` version, never inherited. I think that `__new_stream` would indeed make sense here, then. I'll make that change.

[1] https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name

pescador/mux.py
cjacoby: 👍

pescador/mux.py
cjacoby: Hmm I might need to redo some other things then, since `__new_stream` currently gets the weights in the `BaseMux`. But maybe it shouldn't do that?

pescador/mux.py
cjacoby: On this, I'd say see core.py/`Streamer`, since that behavior is roughly speaking inherited from the base `Streamer` class.

Basically, `activate`/`deactivate` are allowing Streamers to be activated with a Context Manger, giving you the benefits of context manager cleanup when the streamer exits.

Re: multiple simultaneous iteration - you mean, towards an `asyncio` / `coroutine` mux?

cjacoby:
I'll wait to finish those last changes till I get your comments.

pescador/mux.py
bmcfee: Ah, I hadn't realized that futurepast wasn't yet released on pypi.  I'll ping Andy and get back to you.

pescador/mux.py
bmcfee: > It's possible that under the "explicit >> implicit" rule, the child class constructors should handle more of this on their own?

I'd be in favor of that as well.



> However I believe the only necessary ones I need to do that are: k, prune_empty_streams, and random_state, which are essentially common to all streamers.

Why is `k` common to all muxen?  It doesn't seem necessary for chain or shuffle.

pescador/mux.py
bmcfee: Yeah, I don't think the base mux should have anything to do with weights.

If we discover that different randomized muxen have shared functionality, it might be worth having an intermediate class (`StochasticMux`) to implement that, but that might also be overkill.

pescador/mux.py
cjacoby: I kept `k` common in order to simplify the underlying implementation as common as possible (and also just to minimize changes from the original Mux in my refactor). 

While strictly speaking `k` is not necessary for those if you were to implement them from scratch for each class, this does allow `activate` in the `BaseMux` to handle a lot of common setup, and I approached this refactor primarily from a polymorphic inheritance perspective.

I'm not sure which perspective is actually simpler. The polymorphic/inheritance implementation is simpler until it's not. However, I've been thinking about the following things as an argument in your direction / specific implementations for each `Mux` subclass:
 * [Calling functions in python is slow](https://ilovesymposia.com/2015/12/10/the-cost-of-a-python-function-call/)
 * What you said below, potentially towards multiprocessing/threading/asyscio cleanness:
> we could potentially isolate generator state to the iterator method, and make it possible to have the mux support multiple simultaneous iteration.

pescador/mux.py
ejhumphrey: what CBJ said.

also, maybe a relevant question is "why do we need the `core.StreamActivator` rather than making `core.Streamer` a context manager directly with `__enter__`/`__exit__` methods? 

Having typed that tho, I think we're starting to get out of scope for this PR.

pescador/mux.py
cjacoby: # 1 (`k` => `k_active`)

👍  I agree. 

# 2 The more I think about this, what you guys are saying is "`BaseMux` should really be an "interface" (a la java), and not a "base class" (a la c++). I think I've been optimizing for making the minimal possible changes from the original code (also why I didn't rename `k` before), and I need to stop doing that.

pescador/mux.py
cjacoby: I think it is more complicated than that; we don't want the user to have to use the streamer in a context manager; it's a part of internal state.

Anyway - agreed out-of-scope. Perhaps that discussion should get bounced to a separate Issue? It is definitely not a Mux-specific thing, which this PR is.

cjacoby:
> Now, I'm not sure whether this is more wrench or lightbulb, but I think this is an interesting lens to consider this discussion. Consider PoissonMux vs ShuffledMux. I think if you did PoissonMux(k_active=len(streamers), rate=np.inf), you'd get a ShuffledMux, right?

This is in fact exactly how it is implemented right now. `ShuffledMux` inherits from `PoissonMux` and calls the parent constructor with `k=len(streamers)`.

> So what would happen if we factored out the stream-interruption behavior as a function? For example, I might want to RoundRobin a number of streams, with k_active>1 and a Poisson rate of say 50? This would allow one to preserve epoch ordering over bags of samples, while stochastically subsampling at the same time.

I'm not sure I follow exactly what you're getting at, but if I understand correctly, I have in some sense already done this; `BaseMux` now has this function:
```python
def _n_samples_to_stream(self):
    """Return how many samples to stream for a new streamer. None
    makes an infinite streamer. If the `BaseMux` subclass has a
   `rate` field, it would be returned here. The default - None -
    makes the resulting streamers infinite. (`max_iter`=None)
    """
    return None
```

which you just override to change that behavior. For instance, `PoissonMux`'s version is:
```python
def _n_samples_to_stream(self):
    "Returns rate or none."
    if self.rate is not None:
       return 1 + self.rng.poisson(lam=self.rate)
    else:
       return None
```

Then, when a stream is activated, it uses this function like this (and I believe this creates the situation you are describing):
```python
# Get the number of samples for this streamer.
n_stream = self._n_samples_to_stream()

# instantiate a new streamer
return (self.streamers[idx].iterate(max_iter=n_stream),
             self.weights[idx])
````

**Comment by cjacoby** [comment]
Okay, I'm gonna take a crack at summarizing the proposed changes / what I'm going to attempt to do in my next round of changes, from the above comments. Let me know if I got this right:

1. Simplify `BaseMux` `__init__` arguments to only truly common ones. In my current understanding, I believe this should be:

```python
def __init__(self, streamers, random_state=None, prune_empty_streams=True):
```
(So far, all Mux handle `prune_empty_streams`).

2. `__init__`, `activate`, and `deactivate` should be implemented specifically for each subclass of `BaseMux`, based on that submux's interface needs. 

3. Refactor weights out of `def _activate_stream(self, idx)` and `iterate`. If possible, I would still like to try to keep `iterate()` in the Base class, since it *should* be essentially the same for all muxes, and indeed should define the fundamental behavior of a `Mux`.

I think that `iterate()` would then look approximately like this:
```python
def iterate(self, max_iter=None):
    """Yields items from the mux."""
    if max_iter is None:
        max_iter = np.inf

    with core.StreamActivator(self):
        # Main sampling loop
        n = 0

        while n < max_iter and self._streamers_are_active():  # TODO better function name
                                                              # but this is where a child class can kill
                                                              # the looping if all the weights are 0.
            # Pick a stream from the active set
            idx = self._next_sample_index()

            # Can we sample from it?
            try:
                # Then yield the sample
                yield six.advance_iterator(self.streams_[idx])

                # Increment the sample counter
                n += 1
                self.stream_counts_[idx] += 1

            except StopIteration:
                # Oops, this stream is exhausted.

                # If we're disabling empty seeds, see if this stream
                # produced data at any point; if it didn't, turn it off.
                if (self.prune_empty_streams and
                        self.stream_counts_[idx] == 0):
                    self.valid_streams_[self.stream_idxs_[idx]] = False

                # Call child-class exhausted-stream behavior
                # Picking a new streamer for this slot could happen here, or there could be a separate 
                # function for that.
                self._on_stream_exhausted(idx)

            # If everything has been pruned, kill the while loop
            if not self.valid_streams_.any():
                break
```

That is, unless, @bmcfee, you can make a strong argument for why each subclass should have its own `iterate()` which essentially encapsulates the complete functionality of that Mux. I could see this being a thing if a goal here is to try to make all state contained within `iterate()` for multithreading/async purposes, but I think that that may be out of scope for this PR. (However, I'm willing to consider it, if we can decide that that's The Plan _before_ I tackle this).

pescador/mux.py
```diff
@@ -1,5 +1,57 @@
-#!/usr/bin/env python
-'''Stream multiplexing'''
+'''Stream multiplexing
+
+Defines the interface and several varieties of "Mux". A "Mux" is
+is a Streamer which wraps N other streamers, and at every step yields a
```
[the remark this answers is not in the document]
cjacoby: ✔️

pescador/mux.py
```diff
@@ -1,5 +1,57 @@
...
+sample from one of its sub-streamers.
+
+This module defines the following Mux types:
+
+`PoissonMux`
+
+    A Mux which chooses it's active streams stochastically, and chooses
```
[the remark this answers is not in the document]
cjacoby: ✅

cjacoby:
Okay, I'll give that a think.

Going to try to tie this off this week; got sort of sidetracked with lots
of gigs and traveling and holidays.

Also, I'm in NYC Thur/Fri/Mon/Tue if anyone wants to talk about it in
person.

bmcfee:
re: streamactivator

@cjacoby and I just hashed this out in person.  We came to the conclusion that the `StreamActivator` class is unnecessary, and the context manager functionality can be absorbed directly into `Streamer` and `Mux` as needed.  We went over various issues relating to persisting generator state (should not be a problem) and explicit deactivation, and convinced ourselves that there should not be any memory leaks.

bmcfee:
Note 2: roundrobin mux could have an option to randomize the ordering after each cycle

cjacoby:
@bmcfee Question on the above:

`self.weights = np.atleast_1d(self.weights)`

> not sure atleast_1d is what we want here; it should be exactly 1D

^ what would be your preferred solution for handling this here?

I think my preferred solution is to do the pythonic "Easier to ask for forgiveness than permission" and just remove the line.

cjacoby:
Made some good progress, but not back to a point worth sharing yet.

cjacoby:
Made some good progress, but not back to a point worth sharing yet.

cjacoby:
Hey @bmcfee; made some good progress; got all the tests back to passing with the incorporated recommendations mostly complete.

I have two asks, if you could help me out:
1. The test `test_mux.py:test_critical_mux` is failing, but I'm unclear on what this is really testing and if it should indeed be passing as written anyway. Plz help?

2. More test coverage, especially on new classes. `PoissonMux` / the Deprecated Mux are fairly well covered, but new classes could use more, I think.

To summarize, remaining to complete this PR:
- [ ] I haven't yet done the RoundRobin reorder feature above
- [ ] Significant documentation updates / fixes for new formats. (Different PR or same?)
- [ ] Need to read through and make all the docstrings are up to date with the changes.
- [ ] It's possible that some of my function / member names have gotten less clear; happy for suggestions on this as you review.

Going to spend a couple hours on this tomorrow morning; if you get a bit of time to look at it, I think we could have this done by Monday.

bmcfee:
Thanks for this!

So, reversing my previous recommendation, it seems like futurepast is not going to have a full release any time soon.  We should remove that dep and cook up our own deprecation warnings.  (ugh.)

The critical test that's now failing was introduced to resolve an issue in #80 relating to the semantics of `max_iter`.  It looks like the streamer is now returning `max_iter` samples, rather than stopping when the generator is exhausted. (I.e. it's cycling when it shouldn't.)

cjacoby:
Aaah I see. I took a look at #80 and didn't find the answer in fairly short time; I'll update the comment in the test to be clearer about that, and fix the bug.

bmcfee:
If you wanted to add a better comment to that test in doing so, I wouldn't be mad :grin:

bmcfee:
Otherwise, let's do docs audit (points 2 and 3) as a separate PR.

I'll give point 4 a look tonight and comment in CR.

pescador/mux.py
```diff
@@ -286,11 +335,869 @@ def __new_stream(self, idx):
...
+        # These do not depend on the number of streams, k
+        # This function must be filled out in more detail in a child Mux.
+        self.distribution_ = 1. / self.n_streams * np.ones(self.n_streams)
+        self.valid_streams_ = np.ones(self.n_streams, dtype=bool)
+
+    def deactivate(self):
+        """Reset the Mux state."""
+        self.distribution_ = np.zeros(self.n_streams)
+        self.valid_streams_ = np.zeros(self.n_streams)
+
```
bmcfee: I think we shouldn't have `distribution` here anymore, since the stochastic functionality is factored out of `BaseMux`, right?

pescador/mux.py
```diff
@@ -286,11 +335,869 @@ def __new_stream(self, idx):
...
+            generator;
+
+            If RandomState instance, random_state is the random number
+            generator;
+
+            If None, the random number generator is the RandomState instance
+            used by np.random.
+        """
+        self.streamers = streamers
+        self.n_streams = len(streamers)
```
bmcfee: Do we need to have `n_streams` here?  Can we get away without it?

The reason I ask, is that for something like `ChainMux`, I could imagine a scenario in which the streamers are actually provided as a generator (not a list).  In this case, computing the length would fail.

Yeah, this is an extreme/weird case, but I just want to make sure we don't unnecessarily limit ourselves here.  Of course, we need to know the number of streamers to have a distribution over them for stochastic muxen*, and it's obviously necessary for roundrobin.  Just food for thought.

*actually, you could imagine a stick-breaking process for handling a generator of streamers as well: with probability alpha, select from previously activated streamers, adn wp 1-alpha, pull in the next one from the generator.  It's a weird process, but maybe useful.

pescador/mux.py
```diff
@@ -286,11 +335,869 @@ def __new_stream(self, idx):
...
+            # Otherwise, this one's exhausted.
+            # Set its probability to 0
+            self.stream_weights_[idx] = 0.0
+
+        self.weight_norm_ = np.sum(self.stream_weights_)
+
+    def _n_samples_to_stream(self):
+        "Returns rate or none."
+        if self.rate is not None:
+            return 1 + self.rng.poisson(lam=self.rate)
```
bmcfee: this only applies for poissonmux, right?

pescador/mux.py
```diff
@@ -286,11 +335,869 @@ def __new_stream(self, idx):
...
+        """
+        raise NotImplementedError("_next_sample_index() must be implemented in"
+                                  " a child class.")
+
+
+class WeightedStochasticMux(BaseMux):
+    """A Mux which chooses streams randomly (possibly weighted).
+
+    Expands BaseMux with the following features:
+     * Adds `rate` parameter
```
bmcfee: `rate` is a poisson parameter. I don't think it would be used for, eg, shuffledmux?

pescador/mux.py
cjacoby: I think you are right. I have been having an awkward time working with it, but I didn't step back enough to make that conclusion, but in hindsight, that is definitely the right choice.

pescador/mux.py
cjacoby: So, technically, no.

But, I don't see any reason why the parent class can't just have an `n_streams` that a child class doesn't actually use. I think it is a useful property to have around, even if you don't use it.

pescador/mux.py
cjacoby: I could see that. If that's the case, related questions:

* does the `WeightedStochasticMux` just add weights (and I suppose, relatedly,  `distribution_`)?
* Should `ShuffledMux` even have a `rate` parameter? I suppose it might not need to, since that behavior *can* be accomplished by using the `PoissonMux` anyway, and it would make the `ShuffledMux` interface simpler.

pescador/mux.py
cjacoby: Oh, should have read all of 'em before commenting 🙈

cjacoby:
Question wrt `test_critical_mux`;

the original test is written:
```python
def test_critical_mux():
    # Check on Issue #80
    chars = 'abcde'
    streamers = [pescador.Streamer(x * 5) for x in chars]
    mux = pescador.Mux(streamers, k=len(chars), rate=None,
                       with_replacement=False, revive=True,
                       prune_empty_streams=False, random_state=135)
    samples = mux.iterate(max_iter=1000)
    print(collections.Counter(samples))
```

But I think this should fail; if `revive=False`, I think it should have the behavior you describe (and it does), but with `revive=True`, the streamers should restart. 

In my new implementation (this passes), this should look like this:
```python
@pytest.mark.parametrize('mux_class', [
    functools.partial(pescador.mux.Mux, with_replacement=False, revive=False),
    functools.partial(pescador.mux.PoissonMux, mode="exhaustive")],
    ids=["DeprecatedMux",
         "PoissonMux"])
def test_critical_mux(mux_class):
    """This test checks the following:

    When `max_iter` is specified, the `Mux` should return / complete
    when the input generators are complete, and should not cycle infinitely.

    # Check on Issue #80
    https://github.com/pescadores/pescador/issues/80
    """
    chars = 'abcde'
    n_reps = 5
    streamers = [pescador.Streamer(x * n_reps) for x in chars]
    mux = mux_class(streamers, len(chars), rate=None,
                    prune_empty_streams=False, random_state=135)
    samples = list(mux.iterate(max_iter=1000))
    assert len(collections.Counter(samples)) == len(chars)
    assert len(samples) == len(chars) * n_reps
```

Thoughts / confirmation?

cjacoby:
I'm off for now; here's my updated list:

- [ ] RoundRobin reorder on cycle feature
- [ ] `distribution_` refactor
- [ ] Deprecation updates / fixes

pescador/mux.py
cjacoby: Aah on re-read, I see what you mean about computing the length.

What if it were a property, instead, that knew how to handle that case?

cjacoby:
Update; took a crack at reimplementing the `BaseMux` with only what was necessary to complete your `ChainMux` feature where the `streamers` passed in are generator. [Here's the in progress branch](https://github.com/pescadores/pescador/tree/develop/20171210_mux_refactor__distribution_refactor). This refactor, with `ChainMux` in mind, instead of `PoissonMux`, is making it much clearer to me what is *actually* necessary for `BaseMux` to have, and I think will result in a much much cleaner system overall. (And I'm really liking my `ChainMux` implementation!)

I've finished a new `ChainMux`, but I still need to redo all the other `Mux`es.

bmcfee:
> But I think this should fail. I think it should have the behavior you describe (and it does) if revive=False, but with revive=True, the streamers should restart. (I'm actually confused about why this ever passed...?)

Hrm, let's reason through this:

The semantics of `max_iter=M` are (should be) that the streamer can return up to, but not more than, `M` samples.

The original test setup:

```python
    chars = 'abcde'
    streamers = [pescador.Streamer(x * 5) for x in chars]
    mux = pescador.Mux(streamers, k=len(chars), rate=None,
                       with_replacement=False, revive=True,
                       prune_empty_streams=False, random_state=135)
    samples = mux.iterate(max_iter=1000)
```
- 5 streamers, each of which generate a string of length 25.
- All streamers are activated simultaneously, and are sampled without replacement.  This should behave like a shufflemux.
- There is no `rate`, so each streamer runs to completion.

What I would expect, when `revive=False`, is a sequence of length 5*25=125.  Since this is less than `max_iter=1000`, the mux should stop after 125.

When `revive=True`, the sequence should be unbounded in length, so `max_iter=1000` should stop it after making 1000 samples.


> assert len(collections.Counter(samples)) == len(chars)

I think this should be `n_reps` not `len(chars)`?  They happen to be equal in this case (which is bad, suggest changing to `n_reps=7` or some other number relatively prime to 5).

cjacoby:
@bmcfee I agree with your assessment above. 👍

cjacoby:
@bmcfee Question:

Is the `RoundRobinMux` optional shuffle at the end of the cycle not exactly equivalent to a `ShuffleMux`? Or (I can't remember) were you thinking of some other behavior that doesn't match quite exactly?

And, if it *is* the same, does it make sense to actually only have the `RoundRobinMux` with optional shuffle, and no `ShuffleMux`? I think this might be cleaner (implementation-wise) than trying to derive `ShuffleMux` from `WeightedStochasticMux`.

(Although not necessarily cleaner user-wise. `ShuffledMux` could just be a thin wrapper on `RoundRobinMux`, where that defaults to true; I'm not sure what the best thing here is. I think from the end-user standpoint, having a separate class that enacts that behavior is better.)

bmcfee:
> Is the RoundRobinMux optional shuffle at the end of the cycle not exactly equivalent to a ShuffleMux? Or (I can't remember) were you thinking of some other behavior that doesn't match quite exactly?

`ShuffleMux` has the following logic:
```
while True:
   pick streamer i uniformly at random
   yield example from i
```
`RoundRobinMux` has the following:
```
stream_perm = streamers[random permutation]  // not sure if this part is currently included, or optional
while True:
    for i in len(stream_perm):
        yield example from stream_perm[i]
```
I think the proposed end-of-cycle shuffle idea was to wrap the above logic in another loop that restarts all streamers only after they've all been exhausted, and then computes a new permutation of the streamers for the round-robinning.  I don't think this is equivalent to `ShuffleMux`, since there are conditional dependencies to the sampling order within blocks (ie between exhaustions) that `ShuffleMux` does not have.

If that's too complex to implement right now, I'm okay with dropping it and keeping the same permtuation for all iterations of the `RoundRobin`.  This might actually be what most people expect/want anyway.

cjacoby:
I actually just came to the same conclusion while trying to determine
exactly how a `RoundRobinMux` should handle exhaustion behaviors.

Agreed on all the above. I think I probably can still implement this
feature simply, as I do already have the list of streamers anyway, so it
should be trivial.

cjacoby:
Mux feature idea:

Right now a `Mux` only really works with `pescador.Streamer`s as arguments to `streamers`, but what if for convenience they automatically wrapped iterables in streamers if they were not already streamers?

(Particularly thinking about this wrt various tests where there are lots of `pescador.Streamer`s around)

bmcfee:
I'm fine with that, but it's out of scope for this PR.

cjacoby:
I agree, I just needed to write it down somewhere while I was thinking
about it.

cjacoby:
Update:

- [ ] Refactor `BaseMux` to have *only* the necessary components, and no cruft.
    - [ ] `distribution_` refactor to only `PoissonMux`
    - [ ] Fallout/cleanup from the above refactors
        - [ ] Finish updated `PoissonMux`
        - [ ] Finish updated `ShuffledMux`
        - [x] Finish updated `RoundRobinMux`
        - [x] Finish updated `ChainMux`
- [x] RoundRobin reorder on cycle feature
- [ ] Deprecation updates / fixes

cjacoby:
Alright, @bmcfee; that's all of it, minus the deprecations. I think you'll like this better; it's much closer to what I think you've been getting at that I had been in those intermediate stages.

I ended up removing the intermediate class for the PoissonMux and ShuffledMux. I started this iteration on the refactor with `ChainMux` instead of `PoissonMux`, which I think was an excellent plan, as it caused me to reconsider what was actually necessary for the `BaseMux` to have to something much simpler.

Anyway, would appreciate comments on this one; relatively major changes. But, so close now, I think!

bmcfee:
This is looking really good!  I'm a little wary of the `n_streams` property trapping exceptions and returning `None`, but otherwise don't have an opinions right now.  I like flattening the class hierarchy as well, even if it adds a little redundancy.

cjacoby:
Another possible option:

`BaseMux` implements the trivial property for `n_streams`, and `ChainMux`
overrides it with something else (since almost all `Mux` should want the
Base case)? I'm not sure what exactly to have it do in the
streams-are-an-iterator case.

bmcfee:
Or we just drop the idea of supporting chaining on iterators and make it explicit.  I'm fine with that.

cjacoby:
yeah, but i already implemented that thing and it's so shiny... ;)

cjacoby:
Oh! Why didn't I think of this before?

`n_streams` can just return the length of the streamer; `ChainMux` never uses it! (Unless an end user calls `n_streams` on their `ChainMux`, but if they already passed in a generator, they should know not to do that)

cjacoby:
I've done a first pass reading over all the docstrings; I've run some things manually to spot check; I've dealt with some unexpected DeprecationWarnings that came up.

Pending further review, I'm calling this done. 🎉 🎈

cjacoby:
@bmcfee; How do you dig deeper on `depencency-ci` to figure out what error occurred? I can't find a button to give me anything more verbose. I made very minor changes on this push vs the last one which worked, which (looking at the diff) had nothing to do with imports / deps.

bmcfee:
> How do you dig deeper on depencency-ci to figure out what error occurred?

I don't know what happened -- seemed to be a service outage, and nothing related to an error in the PR.  I reran the dependency build and it seems happy again.

bmcfee:
Otherwise, is this ready for a final CR?

cjacoby:
Unless I'm forgetting something, I think so! I've done a couple passes
through and I think it's in a good spot.

bmcfee:
@cjacoby cool, I'll look this afternoon then.

pescador/mux.py
```diff
@@ -286,11 +342,867 @@ def __new_stream(self, idx):
...
+        streamers : iterable of streamers
+            The collection of streamer-type objects
+
+        k_active : int > 0
+            The number of streams to keep active at any time.
+
+        rate : float > 0 or None
+            Rate parameter for the Poisson distribution governing sample counts
+            for individual streams.
+            If ``None``, sample infinitely from each stream.
```
bmcfee: Do we want to persist this behavior (None = inf)?  Or should we start forcing poissonmux users to specify a rate, and use shufflemux if they want unbounded streaming?

bmcfee: Thinking a bit more carefully: shufflemux doesn't support the working subset idea.  The use-case where Poisson with `rate=None` makes sense is where you have streamers that self-limit the number of samples they generate.  It's not really poisson-rated anymore, but it's also not shuffling.  Hmmmm....

pescador/mux.py
```diff
@@ -286,11 +342,867 @@ def __new_stream(self, idx):
...
+        # Initialize each active stream.
+        for idx in range(self.n_streams):
+            # Setup a new streamer at this index.
+            self._new_stream(idx)
+
+    def _streamers_available(self):
+        """As we are treating `streamers` as a generator, we can only know
+        if it is available if the streamer exited or not.
+        """
+        return any([x is not None for x in self.streams_])
```
bmcfee: This will exception after being deactivated because `self.streams_ = None` is not iterable.  Any ideas?

pescador/mux.py
```diff
@@ -286,11 +342,867 @@ def __new_stream(self, idx):
...
+                self._setup_streams(permute=False)
+
+            elif self.mode == "permuted_cycle":
+                self._setup_streams(permute=True)
+
+
+class ChainMux(BaseMux):
+    """As in itertools.chain(). Runs the first streamer to exhaustion,
+    then the second, then the third, etc. k=1.
+
```
bmcfee: what is k=1 here?

pescador/mux.py
```diff
@@ -286,11 +342,867 @@ def __new_stream(self, idx):
...
+        streamers : list of pescador.Streamers OR generator of
+            pescador.Streamrers
+
+        mode : ["exhaustive", "with_replacement"]
+            `exhaustive`
+                `ChainMux will exit after each stream has been exhausted.
+
+            `with_replacement`
+                `ChainMux will restart from the beginning after each
+                streamer has been run to exhaustion.
```
bmcfee: `with_replacement` here means something different than in other muxen.  Maybe `cycle` is a better term?  Or am I misreading?

bmcfee:
@cjacoby looking good!  I had a couple of minor comments to clear up before merging, but I'm pretty happy with it overall.

pescador/mux.py
cjacoby: I don't have a good answer here, but what I've been thinking about is basically:

* `ShuffledMux` shouldn't have too complicated an interface, since it is (not exactly, but more or less) a subset of what `PoissonMux` can do. I'm thinking of it as a shortcut for a particular use case.

* Perhaps, to your point, it might be worthwhile to make `rate` a required parameter (no default), but still allow `np.inf` (or even, perhaps, None) if the user actually wants to allow unbounded streaming?

pescador/mux.py
cjacoby: * The comment here is copypasta. Whoops. I fix.

* `_streamers_available()` only gets called in `iterate()` after activation, and is not useful outside of that. So, I don't see that this is a problem...?

pescador/mux.py
cjacoby: mrp old; my eyes skimmed over it without seeing! getting rid of it.

pescador/mux.py
cjacoby: Agreed; this felt a little weird to me, but I was (at the time) trying to have the behavior have the same mode name across classes. I think `cycle` would be good, and would match `RoundRobinMux`.

pescador/mux.py
cjacoby: OTOH, `rate` defaults to 256, and if you want it to be something other than that you have to specify it by default anyway.

I think having the `None` feature is very useful, even if it is not strictly Poisson.

pescador/mux.py
bmcfee: Ok, cool.  I guess I missed that it was a private method (such as they are in python), so I'm fine with that.

pescador/mux.py
bmcfee: Yeah, I agree with keeping the `None` behavior.  Also I think making `rate` a positional/required argument makes a lot of sense.

cjacoby: Okay, I'll do that.

cjacoby:
Okay, those are all the requested changes. Are we good?

bmcfee:
I think so!  I'll let the CI finish out, and then merge.  A+

Issue #108, opened 2017-12-15
Remove StreamActivator class


bmcfee:
We removed the stream activator logic from `mux` in #103, but forgot to do the same for general Streamers.  See the discussion thread in #103 for why this makes sense and how to do it.

cjacoby:
(That's not exactly true. We just talked about it but you said not to do it
for that PR.)

bmcfee:
> (That's not exactly true. We just talked about it but you said not to do it
> for that PR.)

Yeah, you're right.  Brain is going in my old age :rofl:.  But we should do it now / after merging #109

cjacoby:
It is done

PR #111, opened 2017-12-18
Remove StreamActivator class

cjacoby:
Implements #108 - Removes separate `StreamActivator` class, and converts `Streamer`  to using itself as a Context Manager instead. (for both `Streamer` and `BaseMux`).

### DIFF
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -7,21 +7,6 @@ import six
 from .exceptions import PescadorError
 
 
-class StreamActivator(object):
-    def __init__(self, streamer):
-        if not isinstance(streamer, Streamer):
-            raise PescadorError("`streamer` must be / inherit from Streamer")
-        self.streamer = streamer
-
-    def __enter__(self, *args, **kwargs):
-        self.streamer.activate()
-        return self
-
-    def __exit__(self, *exc):
-        self.streamer.deactivate()
-        return False
-
-
 class Streamer(object):
     '''A wrapper class for recycling iterables and generator functions, i.e.
     streamers.
@@ -107,6 +92,14 @@ class Streamer(object):
         self.kwargs = kwargs
         self.stream_ = None
 
+    def __enter__(self, *args, **kwargs):
+        self.activate()
+        return self
+
+    def __exit__(self, *exc):
+        self.deactivate()
+        return False
+
     @property
     def active(self):
         """Returns true if the stream is active
@@ -145,7 +138,8 @@ class Streamer(object):
         cycle : force an infinite stream.
 
         '''
-        with StreamActivator(self):
+        # Use self as context manager / calls __enter__() => activate()
+        with self:
             for n, obj in enumerate(self.stream_):
                 if max_iter is not None and n >= max_iter:
                     break
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -93,11 +93,11 @@ class Streamer(object):
         self.stream_ = None
 
     def __enter__(self, *args, **kwargs):
-        self.activate()
+        self._activate()
         return self
 
     def __exit__(self, *exc):
-        self.deactivate()
+        self._deactivate()
         return False
 
     @property
@@ -107,7 +107,7 @@ class Streamer(object):
         """
         return self.stream_ is not None
 
-    def activate(self):
+    def _activate(self):
         """Activates the stream."""
         if six.callable(self.streamer):
             # If it's a function, create the stream.
@@ -117,7 +117,7 @@ class Streamer(object):
             # If it's iterable, use it directly.
             self.stream_ = iter(self.streamer)
 
-    def deactivate(self):
+    def _deactivate(self):
         self.stream_ = None
 
     def iterate(self, max_iter=None):
@@ -138,7 +138,7 @@ class Streamer(object):
         cycle : force an infinite stream.
 
         '''
-        # Use self as context manager / calls __enter__() => activate()
+        # Use self as context manager / calls __enter__() => _activate()
         with self:
             for n, obj in enumerate(self.stream_):
                 if max_iter is not None and n >= max_iter:
--- a/pescador/mux.py
+++ b/pescador/mux.py
@@ -214,7 +214,8 @@ class Mux(core.Streamer):
         self.weight_norm_ = None
 
     def iterate(self, max_iter=None):
-        with core.StreamActivator(self):
+        # Calls Streamer's __enter__, which calls activate()
+        with self:
 
             # Main sampling loop
             n = 0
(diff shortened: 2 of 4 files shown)


pescador/core.py
```diff
@@ -107,6 +92,14 @@ def __init__(self, streamer, *args, **kwargs):
         self.kwargs = kwargs
         self.stream_ = None
 
+    def __enter__(self, *args, **kwargs):
+        self.activate()
+        return self
```
bmcfee: This all seems good to me.  One question: do we want to keep `activate` as a method?  It's only ever called from `__enter__`, and that method never seems to change.  Seems like these could be consolidated.

OTOH, `deactivate` does double-duty as a reset operation, and gets called both in `init` and `__exit__`, so they couldn't be consolidated directly.

I bring this up because it seems like a bit of redundancy here that we could do away with at this point.

If nothing else, shouldn't the activate/deactivate methods be `_(de)activate` to discourage calling from outside?

bmcfee:
Maybe OT, but do you know how the activation / context management interacts with muxing with replacement (eg, multiple activation of the same streamer)?

pescador/core.py
cjacoby: I feel torn about keeping `(de)activate`. I thought about removing them, and figured I would let it come to discussion here.

Currently, my feeling is that because of this:
> deactivate does double-duty as a reset operation, and gets called both in init and __exit__, so they couldn't be consolidated directly.
it is better to keep them explicit. Also I like children overriding `(de)activate` a little better than overriding `__enter__/__exit__`.  But OTOH I agree that it might be redundant / unnecessary.

Also, yes, you're right; they should be `_(de)activate`. I'll fix that at least, until we can come up with a conclusion.

cjacoby:
> Maybe OT, but do you know how the activation / context management interacts with muxing with replacement (eg, multiple activation of the same streamer)?

Hmm. Not sure. I could see this going either way:
1. when a new active stream is created via .iterate(), it returns a new generator with fresh state? Somehow I doubt this is what is actually happening.
2. when a new active stream is created via .iterate(), they share the state, and we have a problem; the state for each iterator *must* be separate from each other one. I suspect now that you mention it that this is happening.

This should be easily testable, no? Something like:

```python
def test_muxing_with_replacement():
    streamers = [pescador.Streamer(_cycle, 'a'),
                          pescador.Streamer(_cycle, 'b')]
    # Is it actually valid to have k_active > n_streamers? Should it be?
    # ... But I think that would test it, if it is.
    mux = pescador.PoissonMux(streamers, 4, rate=20)
    ....
```

If this is the case, preferred solution? (Feel free to punt this discussion to a new Issue).

bmcfee:
> when a new active stream is created via .iterate(), they share the state, and we have a problem; the state for each iterator must be separate from each other one. I suspect now that you mention it that this is happening.

Confirmed: probing the hash of the iterators when sampling with replacement reveals that they're the same object.

```python
In [2]: S = pescador.Streamer('a')

In [3]: M = pescador.PoissonMux([S, S], k=5, mode='with_replacement')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-6eda7189c301> in <module>()
----> 1 M = pescador.PoissonMux([S, S], k=5, mode='with_replacement')

TypeError: __init__() got an unexpected keyword argument 'k'

In [4]: M = pescador.PoissonMux([S, S], 5, rate=None, mode='with_replacement')

In [5]: xes = M.iterate()

In [6]: M.streams_

In [7]: next(xes)
Out[7]: 'a'

In [8]: M.streams_
Out[8]: 
[<generator object Streamer.iterate at 0x7ff2f23a2308>,
 <generator object Streamer.iterate at 0x7ff2f23a2a98>,
 <generator object Streamer.iterate at 0x7ff2e8151eb8>,
 <generator object Streamer.iterate at 0x7ff2db7a2d00>,
 <generator object Streamer.iterate at 0x7ff2db7a2888>]

In [9]: M.streamers
Out[9]: 
[<pescador.core.Streamer at 0x7ff2e80b85c0>,
 <pescador.core.Streamer at 0x7ff2e80b85c0>]

In [10]: M.streamers[0]
Out[10]: <pescador.core.Streamer at 0x7ff2e80b85c0>

In [11]: x1 = M.streamers[0]

In [13]: x2 = M.streamers[1]

In [14]: x1.stream_
Out[14]: <str_iterator at 0x7ff2e8026ef0>

In [15]: x2.stream_
Out[15]: <str_iterator at 0x7ff2e8026ef0>
```

This becomes a problem because you have two active streamers that are now mutually dependent.  This is unavoidable as long as we're manipulating the streamer object's state within the context manager's `__enter__`.  If, however, we have the context manager make a copy of the streamer to activate, this probably won't be a problem?

cjacoby:
I agree, I think.

Is this a concern that should potentially apply to all Mux? Or just
PoissonMux? If it is, I'd rather do it now while my head is still in this
code.

One possible solution, as I'm pondering: an `iterate()` helper from
`Streamer` which returns a generator from a copy of self? Perhaps
`iterate(copy=True)`, or something like that? Then, `PoissonMux` can always
just ask it for the copy.

cjacoby:
Nvm I just re-read your comment and I think I understand what you were
saying better; I agree that the contextmanager is the spot to do this.

bmcfee:
I think this potentially affects all muxen, eg, when providing multiple references to the same streamer as inputs.  I wouldn't normally do that, but I think it's fair to expect each streamer (reference) to behave independently.  Where this gets really nasty is when stacking muxen, which could have deep state to copy.

I think returning a (shallow) copy of self in the context manager is sane, but I'm not 100% sure of that yet.  We'll need to rewrite things a bit so that the iterate loop talks to the context instead of `self`, but that shouldn't be too bad (i think).

cjacoby:
I agree that that shouldn't be too bad in theory ;).

Q: is this problem something to solve in this PR or another one?

I propose we close this off more or less as is (with the minor requested
changes), and I'll start a new issue for the mux copying.

bmcfee:
> propose we close this off more or less as is (with the minor requested
> changes), and I'll start a new issue for the mux copying.

I'm okay with that.

cjacoby:
I tried to add a test to fix the coveralls issue, but it didn't work. I'm inclined to not worry about it, since the "coverage loss" is not due to an increase in code.

bmcfee:
It looks like there's a coverage gap due to shuffledmux never deactivating in the tests?  Should it deactivate on init?

cjacoby:
Indeed; you were right, because I missed changing `ShuffledMux`'s `deactivate()` to `_deactivate()`. `_deactivate()` actually gets called in the `BaseMux.__init__()`. That should do it...

Issue #112, opened 2017-12-19
Make `Streamers`' `iterate()` return a copy of the streamer


cjacoby:
As identified in #111, when activated, each `Streamer` generator maintains the state of the original object. This is a bug in situations such as `PoissonMux(mode='with_replacement')`, where the stream can be used multiple times. (See discussion in #111 for more details).

Proposals for fixing:
* Have the context manager make a copy of the streamer to activate, which is then used instead of `self` within the context.

Questions / Clarifications:
* Is returning a shallow copy sufficient?
* Option for deep copy? (i.e. `deep=True/False`)
* Maybe provide an option in `.iterate()` to set whether a copy happens or not, for the following case identified by @bmcfee:
> ... I think it's fair to expect each streamer (reference) to behave independently. *Where this gets really nasty is when stacking muxen, which could have deep state to copy.*

bmcfee:
> Is returning a shallow copy sufficient?

Nope: if the streamers are muxes, there's non-trivial state to propagate.  If they're muxes of muxes, a shallow copy would fail.

> Option for deep copy? (i.e. deep=True/False)

I'd prefer to limit options as much as possible.  These objects are generally lightweight, so a deepcopy shouldn't be too bad?

bmcfee:
One thought though: returning a copy of the object as a context might play tricks with any callback / state logging we want to do #94 .  @ejhumphrey thoughts?

cjacoby:
Fixed in #113.

Issue #110, opened 2017-12-16
Mux Feature - Accept iterables (instead of just Streamers), and wrap them in Streamers automatically


cjacoby:
There are many places in the Mux tests where this happens:

```python
>>> a = pescador.Streamer('a')
>>> b = pescador.Streamer('b')
>>> c = pescador.Streamer('c')

>>> mux = pescador.RoundRobinMux([a, b, c])
>>> "".join(list(mux.iterate()))
abc
```

I believe this could be easily simplified by allowing all `Mux` to accept iterables, and just automatically wrapping them in a `pescador.Streamer`, and documenting that this is what will happen if you use it this way. A convenience method, of sorts. That would make the above code possibly:

```python
>>> mux = pescador.RoundRobinMux(['a', 'b', 'c'])
>>> "".join(list(mux.iterate()))
abc
```

(a trivial example, but I think it gets the point across)

(Discussed briefly in #103)

bmcfee:
:+1: from me on this. @ejhumphrey ?

bmcfee:
This seems like a one-liner change in the base class `__init__` method:

```python
self.streamers = streamers
```
becomes
```python
self.streamers = [s if isinstance(s, Streamer) else Streamer(s) for s in streamers]
```

This feels consistent with the `Streamer` base-class behavior of handling iterables and generators in `_activate()`.

Issue #94, opened 2017-05-05
Mux State Interface


ejhumphrey:
For reasons that I like and make @bmcfee nervous, I'd like to implement an interface that would allow a user to modify the distribution of the Mux on the fly.

ejhumphrey:
I'll volunteer to take a look at this and determine whether this is in/out for 1.1.

bmcfee:
I think it should probably stay out for 1.1, since 2 will include a pretty substantial refactor of the mux anyway.

ejhumphrey:
punt on this for 1.1, deal with on a case by case basis, maybe factor common interfaces out later should they (actually) materialize

bmcfee:
Circling back since the mux refactor is finshed #107.

@ejhumphrey do you have concrete ideas for what exactly you want to do to mux states?  I'm still wary about allowing external control, since that would really mess up things like chainmux or roundrobin.

bmcfee:
Re-upping for the new year: @ejhumphrey ?

ejhumphrey:
Thanks for the poke ... if I'm not mistaken, based on my peripheral awareness of what's been happening here, I'll need to set aside some time to look into the current state of master to respond intelligently. I'll throw this on the backlog, but it's looking like I can get to this at the end of the week (couple Weds deadlines).

cjacoby:
I'm in NYC this week if you need some guidance; I've got all this solidly
in my brain right now.

bmcfee:
I'm wondering if it makes sense to punt this up to 2.1, and scoping 2.0 to just be the mux refactor (and requisite other updates)?

cjacoby:
I don't have a good sense of what is trying to be achieved with this one. Unlike #116 and #110, I think this one has a slightly higher probability of affecting the API, so it might be good to at least get an "on-paper" proposal in this thread for exactly what this might mean, before deciding that?

-C

bmcfee:
Fair enough.  Pinging @ejhumphrey again.

bmcfee:
Reflecting on this a bit more: the one big missing piece of this is a way to track which iterates came from which streamers, so that you have a way to measure the loss per streamer and derive weight updates accordingly.  I see two options here:

1. Customize your streamers to include metadata as part of their output stream
2. Adapt the mux to generate the metadata for you

Option 2 makes me nervous because it's fundamentally incompatible with the core `Streamer` API -- why would a single streamer ever need to do that?

[...]

PR #113, opened 2017-12-20
Fix Copy Streamers Bug

cjacoby:
Implements #112.

When activated, streamers now return a copy of themselves. This is necessary in order to allow a single streamer to be used multiple times. The most obvious use case here (which did not work before!) was the `PoissonMux` in `with_replacement` mode, where a single stream should be able to be used multiple times.

In order to achieve this, each streamer (both `pescador.mux.BaseMux` derived classes and `pescador.Streamer`) operates on a new, deep-copy of itself each time it is activated with `.iterate()`.

Includes various updates to tests to test for this behavior.

### DIFF
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -88,9 +88,18 @@ class Streamer(object):
             raise PescadorError('`streamer` must be an iterable or callable '
                                 'function that returns an iterable object.')
 
+        # The iterable or callable to stream from
         self.streamer = streamer
+        # Args and kwargs are passed to an instantiated function
         self.args = args
         self.kwargs = kwargs
+
+        # When a stream is activated, a copy of this streamer is made.
+        # The number of copies is tracked with active_count_.
+        self.active_count_ = 0
+
+        # Stream points to the activated generator. This is only used
+        # in the copy created.
         self.stream_ = None
 
     def copy(self):
@@ -99,18 +108,30 @@ class Streamer(object):
     def __enter__(self, *args, **kwargs):
         streamer_copy = self.copy()
         streamer_copy._activate()
+
+        # Increment the count of active streams.
+        self.active_count_ += 1
+
         return streamer_copy
 
     def __exit__(self, *exc):
         self._deactivate()
+
+        # Decrement the count of active streams.
+        self.active_count_ -= 1
+
+        if self.active_count_ < 0:
+            raise PescadorError("Active stream count passed below 0 for {}"
+                                .format(self))
+
         return False
 
     @property
     def active(self):
         """Returns true if the stream is active
-        (ie a StopIteration) has not been thrown.
+        (ie there are still open / existing streams)
         """
-        return self.stream_ is not None
+        return self.active_count_
 
     def _activate(self):
         """Activates the stream."""
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 """Core classes"""
 import collections
+import copy
 import inspect
 import six
 
@@ -92,9 +93,13 @@ class Streamer(object):
         self.kwargs = kwargs
         self.stream_ = None
 
+    def copy(self):
+        return copy.deepcopy(self)
+
     def __enter__(self, *args, **kwargs):
-        self._activate()
-        return self
+        streamer_copy = self.copy()
+        streamer_copy._activate()
+        return streamer_copy
 
     def __exit__(self, *exc):
         self._deactivate()
@@ -139,8 +144,8 @@ class Streamer(object):
 
         '''
         # Use self as context manager / calls __enter__() => _activate()
-        with self:
-            for n, obj in enumerate(self.stream_):
+        with self as active_streamer:
+            for n, obj in enumerate(active_streamer.stream_):
                 if max_iter is not None and n >= max_iter:
                     break
                 yield obj
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -106,23 +106,33 @@ class Streamer(object):
         return copy.deepcopy(self)
 
     def __enter__(self, *args, **kwargs):
-        streamer_copy = self.copy()
-        streamer_copy._activate()
+        # If this is the base / original streamer, create a copy and return
+        # it
+        if not self.is_activated_copy:
+            streamer_copy = self.copy()
+            streamer_copy._activate()
 
-        # Increment the count of active streams.
-        self.active_count_ += 1
+            # Increment the count of active streams.
+            self.active_count_ += 1
+
+        # However, if this is an activated streamer, then it is a copy,
+        #  so just return self.
+        else:
+            streamer_copy = self
 
         return streamer_copy
 
     def __exit__(self, *exc):
-        self._deactivate()
+        if not self.is_activated_copy:
 
-        # Decrement the count of active streams.
-        self.active_count_ -= 1
+            self._deactivate()
 
-        if self.active_count_ < 0:
-            raise PescadorError("Active stream count passed below 0 for {}"
-                                .format(self))
+            # Decrement the count of active streams.
+            self.active_count_ -= 1
+
+            if self.active_count_ < 0:
+                raise PescadorError("Active stream count passed below 0 for {}"
+                                    .format(self))
 
         return False
 
@@ -132,6 +142,13 @@ class Streamer(object):
         (ie there are still open / existing streams)
         """
         return self.active_count_
+
+    @property
+    def is_activated_copy(self):
+        """is_active is true if this object is a copy of the original Streamer
+        *and* has been activated.
+        """
+        return self.stream_ is not None
 
     def _activate(self):
         """Activates the stream."""
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -90,6 +90,7 @@ class Streamer(object):
 
         # The iterable or callable to stream from
         self.streamer = streamer
+
         # Args and kwargs are passed to an instantiated function
         self.args = args
         self.kwargs = kwargs
@@ -118,8 +119,8 @@ class Streamer(object):
         return copy_result
 
     def __enter__(self, *args, **kwargs):
-        # If this is the base / original streamer, create a copy and return
-        # it
+        # If this is the base / original streamer,
+        #  create a copy and return it
         if not self.is_activated_copy:
             streamer_copy = copy.deepcopy(self)
             streamer_copy._activate()
@@ -127,7 +128,7 @@ class Streamer(object):
             # Increment the count of active streams.
             self.active_count_ += 1
 
-        # However, if this is an activated streamer, then it is a copy,
+        # However, if this is an "activated" streamer, then it is a copy,
         #  so just return self.
         else:
             streamer_copy = self
@@ -136,8 +137,6 @@ class Streamer(object):
 
     def __exit__(self, *exc):
         if not self.is_activated_copy:
-
-            self._deactivate()
 
             # Decrement the count of active streams.
             self.active_count_ -= 1
@@ -171,9 +170,6 @@ class Streamer(object):
         else:
             # If it's iterable, use it directly.
             self.stream_ = iter(self.streamer)
-
-    def _deactivate(self):
-        self.stream_ = None
 
     def iterate(self, max_iter=None):
         '''Instantiate an iterator.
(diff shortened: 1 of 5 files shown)


cjacoby:
Hey @bmcfee! Happy holidays ;).

## Concerns / problems identified
In debugging `ChainMux` bugs which fall out of the `Streamer` changes I made, I have identified the following issues, which I hesitate to take action on without some discussion:

1. `ChainMux` (in my implementation) makes use of the `.active` property of `Streamer`. This feature is not actually tested in `test_core.py`, I learned!
2. Making `__enter__()` return a copy of `self` works great! Except that it completely breaks the ability to track whether the streamer is still `.active`, because there is no reference to the now-active streamer, except within the ContextManager.

## Discussion
I'm mostly concerned with how to solve Concern 2 above, since how it is solved will determine the solution to Concern 1.

One solution:
Add a list of references
a`Streamer.__enter__()`

## Proposals
0. Add tests for `.active` in `core.py`. 
1.

bmcfee:
Oof, this is a doozy.  

> I'm mostly concerned with how to solve Concern 2 above, since how it is solved will determine the solution to Concern 1.
> 
> One solution:
> Add a list of references to child streamers / active streamers, and add the copy to them in __enter__(). Then remove it in __exit__.
> 

So this PR generally turns streamers into factories (yuck, i know), and then the factory has to keep track of / collect everything it constructs?  I guess that's not so bad if the context manager is the only interface to instantiating streams, but it somehow feels a bit awkward.


> .active would then just turn len(active_streamers)
> 
> I think that the .active property is useful, and that maintaining the references is nice for being able to introspect what is happening with the streamer later (#94).

I think I agree, though the multiple instantiation thing makes having a `bool`-type active property less appealing.  I like your proposal to return the number of active streams instead -- basically turning a semaphore into a counting semaphore. :+1:


Thinking a bit more about this comment:

> Making __enter__() return a copy of self works great! Except that it completely breaks the ability to track whether the streamer is still .active, because there is no reference to the now-active streamer, except within the ContextManager.

Do we even need references here?  What if `__enter__` just increments a counter, and `__exit__` decrements it?  (Throw an exception if it's gonna dip negative.)  Then the `active` property can just return the current count?  Or is the problem that `__exit__` is called on the copy, not the master stream?  Maybe a parent reference could be useful here?


---


> 
> I'm also pondering about some things that falls out of this:
> 
> You might want to have the "child copies" have a flag that lets them know they are copies, so they can't make new copies.

Why is that bad?  When would it ever happen anyway?  You'd have to do something like:

```python
with stream as ctx:
   with ctx as ctx2:
      ...
```
which seems really unlikely, if not totally preventable within mux internals?

> ... or, the child copies actually should be a different class. But, this seems like it is headed towards unnecessary complication, and i don't actually want to go down this route.

Agreed.

> Another solution - easier, but I like it less:
> 
> Just remove .active, and similarly, simplify ChainMux so that it can't accept an iterable (which requires the Streamer.active)

That would fix chainmux, and I'm okay with that.  I think it still leaves us with the problem of supporting #94 downstream though?

cjacoby:
> So this PR generally turns streamers into factories (yuck, i know), and then the factory has to keep track of / collect everything it constructs? I guess that's not so bad if the context manager is the only interface to instantiating streams, but it somehow feels a bit awkward.

Yep, what you said. 

This is all a bit messier / more complicated than ideal, but I don't really see another path, currently, given the `with_replacement` requirements we've established. 

So, if I understand correctly, I'm going to consider this a go-ahead on the following:
- [ ] `__enter__` / `__exit__` increment/decrement a stream counter
- [ ] `.active` now just returns the active stream counter, and you can treat it as a bool. (I think `.n_active` might be a better name, now, though)

I think that's all that should be necessary to fix the `ChainMux`, and actually keeps this relatively simple / straightforward (at least, until we find weird Mux cases)

bmcfee:
Reading the python core docs a bit, it seems that there's precedent for this copy-on-context pattern: [decimal.localcontext](https://docs.python.org/3/library/stdtypes.html#typecontextmanager).  I'm still a little hazy on which object's `__exit__` function gets called, but I guess it will all work out?

cjacoby:
@bmcfee: After implementing the above, and looking into Mux, I have the following questions/concerns/observations:

1. I think that a base `pescador.Streamer` works very nicely with this "factory" approach and counting it's __enter__/__exits__, creating a new stream for each .iterate(), such that each active stream maintains it's own state.

2. Do we still / definitely want a `Mux` to do this - specifically, allow internal copies and count them?
    - (option A - easy way) I could see keeping it this way, such that you can compose Mux of Mux (which we definitely want to be able to do). To do this, I just need to add an `self.active_count_ = 0` to the `BaseMux` constructor.
    - (option B) However, you could also use `Streamers` to wrap `Mux` to get the same behavior. That would solve the problem more simply (err... ultimately, not right now), I think. This would work well with #110, as each Mux would get wrapped in a Streamer internally.
        - Under the scenario developing here, I think I would actually separate `Mux` from being derived from `pescador.Streamer`. (They could still have a common parent or "interface" in the Java sense which forces them to implement `.iterate()`, and maybe has the `cycle()`, `__call__()` and `__iter__()` common methods).
        - (an aside; I think we could actually remove the context manager from Mux with this change, potentially simplifying it, something you had mentioned wanting to do earlier)

[Random Footnote]
I just realized that with the current changes it would be very easy to have an
`async def aiterate(...)`
Which would be nice for coroutines, etc. :)

cjacoby:
> I'm still a little hazy on which object's __exit__ function gets called, but I guess it will all work out?

I wrote some tests that check this, I think we're good.

bmcfee:
> I think that a base pescador.Streamer works very nicely with this "factory" approach and counting it's enter/exits, creating a new stream for each .iterate(), such that each active stream maintains it's own state.

:+1:



> Do we still / definitely want a Mux to do this - specifically, allow internal copies and count them?

Couldn't parse this -- do you mean to make internal copies of the mux, or internal copies of the activated streamers?

I think in either case, the mux should have the same semantics (wrt copying) as simple streamers.  This keeps the application code clear and easy to read, without requiring users to know why a mux behaves slightly differently and needs special handling.

> (option A - easy way) I could see keeping it this way, such that you can compose Mux of Mux (which we definitely want to be able to do). To do this, I just need to add an self.active_count_ = 0 to the BaseMux constructor.

That seems reasonable to me.  What's the downside?



> I just realized that with the current changes it would be very easy to have an
> async def aiterate(...)
> Which would be nice for coroutines, etc. :)

Yes, but to maintain backwards compat, I think this would have to live in a separate (derived) class and source file that only gets imported if the python version supports async.

cjacoby:
Hey @bmcfee - I'm ready for a code review on this now. I'm going to just go back over my changes and make sure everything is tidy, but I think it's done.

pescador/core.py
```diff
@@ -87,23 +88,65 @@ def __init__(self, streamer, *args, **kwargs):
...
+        # When a stream is activated, a copy of this streamer is made.
+        # The number of copies is tracked with active_count_.
+        self.active_count_ = 0
+
+        # Stream points to the activated generator. This is only used
+        # in the copy created.
         self.stream_ = None
 
+    def copy(self):
+        return copy.deepcopy(self)
```
bmcfee: Should this be implemented as  `__copy__` and `__deepcopy__`?  https://docs.python.org/3/library/copy.html

pescador/mux.py
```diff
@@ -155,9 +155,7 @@ def __init__(self, streamers, k,
 
         self._deactivate()
 
-        if random_state is None:
-            self.rng = np.random
-        elif isinstance(random_state, int):
+        if isinstance(random_state, int) or random_state is None:
             self.rng = np.random.RandomState(seed=random_state)
```
bmcfee: I'm not sure about this change.  Previously, using `random_state=None` would tie directly to the `np.random` state, but now it constructs a new randomstate object.  I don't think these are equivalent, since the former is reproducible with a global seed and the latter is not.  Or am I missing something?

pescador/mux.py
```diff
@@ -215,73 +225,82 @@ def _deactivate(self):
...
                 # Can we sample from it?
                 try:
                     # Then yield the sample
-                    yield six.advance_iterator(self.streams_[idx])
+                    yield six.advance_iterator(active_mux.streams_[idx])
 
                     # Increment the sample counter
                     n += 1
-                    self.stream_counts_[idx] += 1
+                    active_mux.stream_counts_[idx] += 1
```
bmcfee: Where does the stream count decrement?  Or can it only reset to 0 below?

bmcfee:
Looks pretty good -- I haven't made it through the tests yet though.

One question: do you have a handle on what happens when random states are deepcopied within muxen?  I think it should all work out as expected, but want to be sure.

pescador/core.py
cjacoby: err, yes.

pescador/mux.py
cjacoby: I made this change because it is not possible to do a deepcopy on a module object, which is what `self.rng` is if you give it `None` in the previous implementation.

> since the former is reproducible with a global seed and the latter is not. Or am I missing something?

Good point, I had not considered this.

We could solve this another way with a more specific `__deepcopy__` as you mention, where the `__deepcopy__` method knows how to copy the random_state/rng. I'll take a stab at this approach.

pescador/mux.py
cjacoby: `stream_counts_` counts the number of samples drawn from a particular stream to see if it is / is not dead. There is no reason to decrement.

cjacoby:
While making the above __deepcopy__ and rng changes, I will make sure there are tests which check for random states on deepcopied muxen. I also think it should be good, but I'll add the tests to make sure it's explicit, and then ping back for a re-CR / tests-CR.

cjacoby:
Okay @bmcfee, I've more or less got this implemented, but I've hit one small snag I could use some advice on:

When you copy (or deepcopy) a np.random.RandomState, how do you know / prove that they're the same afterwards? There's `RandomState.get_state()`, but I'm not sure how to compare the outputs of `get_state()` to get their equivalence. The googles are not helping me out here.

I'll push my current state so you can see what i've done; for reference, `tests/test_mux.py::TestCopyMux::test_deepcopy__randomseed`.

bmcfee:
> There's RandomState.get_state(), but I'm not sure how to compare the outputs of get_state() to get their equivalence.

Yeah, I'm not sure either.  I would think that a strict tuple equivalence check would work, but might be too strict?

bmcfee:
I just did a quick check, and setting the state (even with different seeds) seems to be sufficient to ensure sample equivalence between randomstate objects.

```python
>>> r1 = np.random.RandomState(seed=100)
>>> r2 = np.random.RandomState(seed=101)
>>> state1 = r1.get_state()
>>> r2.set_state(state1)
>>> sample1 = r1.random_sample(30)
>>> sample2 = r2.random_sample(30)
>>> np.allclose(sample1, sample2)
```

cjacoby:
Okay! Requested changes made; I learned that a deepcopy on a RandomState does indeed successfully copy the internal state as you would expect. (When I had done an initial test in a ipython session before, I wasn't so sure, but my unit tests confirm it).

One last thing on my mind here: I strongly suspect `_deactivate()` is never really being called on the copies. I'm not sure this really matters, perse, but options:
* Move the `_deactivate()` call *out* of the `__exit__` function (where it doesn't actually do anything, really), and put it at the end of the `.iterate()` call, after the loop has exited. 
* Call deactivate in a `__del__`, when the object goes out of scope (because iteration stopped)
* Just remove `._deactivate()` altogether, since it's basically doing nothing right now.
* ... do nothing, leave it as is (albeit possibly leaving dead code).

Thoughts?

Other than that, I'm ready for final CR.

bmcfee:
> Just remove ._deactivate() altogether, since it's basically doing nothing right now.

Is deactivate always a noop?  If so, I'd advocate removing it entirely.  If there is session-specific properties that need to be cleaned up, eg in ZMQStreamer, `__del__` seems like a more natural place anyway.

cjacoby:
I'll throw some debugs in to check, but I'm pretty sure.

cjacoby:
Okay, that should do it. I removed `_deactivate()` from the `Streamer` class, and converted it into `_reset()` in the `Mux` classes, since it is still getting used in the Mux to set variables in the constructor.

bmcfee:
This all looks good to me!  I think we're good to merge.

PR #118, opened 2018-01-21
fixes #116 - cycle now respects max_iter

### LINKED ISSUE #116: max_iter vs cycle?
In implementing the epoch example for #114, I noticed that `cycle` and `max_iter` don't play nicely together.  I guess this was always the case, and it makes sense: cycle should run indefinitely, but max_iter explicitly bounds the number of steps.

I think it's not unreasonable to expect
```python
streamer(cycle=True, max_iter=N)
```
to still generate `N` steps, and cycle internally if necessary to get there.

Is that behavior that we want to support?  The `RoundRobinMux` kind of does this in permuted mode, so it could be generalized.  OTOH, I'm fine with just saying that the two modes are incompatible and throwing an exception if `cycle and max_iter is not None`.

bmcfee:
This PR fixes #116 so that `max_iter` is respected when cycling a Streamer.  This is primarily useful when you want to produce exactly `max_iter` steps, even if the underlying generator stops at `n < max_iter`.

It is also useful for getting some finnicky corner-case behaviors in `PoissonMux` to behave appropriately, eg, epoch sampling.

AFAIK this is ready for CR.

### DIFF
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -220,7 +220,7 @@ class Streamer(object):
 
         count = 0
         while True:
-            for obj in self:
+            for obj in self.iterate():
                 count += 1
                 if max_iter is not None and count > max_iter:
                     return
--- a/pescador/core.py
+++ b/pescador/core.py
@@ -55,6 +55,11 @@ class Streamer(object):
     >>> for i in stream.cycle():
     ...     print(i)  # Displays 0, 1, 2, 3, 4, 0, 1, 2, ...
 
+    Or finitely many examples, restarting the generator as needed
+
+    >>> for i in stream.cycle(max_iter=7):
+    ...     print(i)  # Displays 0, 1, 2, 3, 4, 0, 1
+
 
     An alternate interface for the same:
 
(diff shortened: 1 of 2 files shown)


pescador/core.py
```diff
@@ -196,19 +201,29 @@ def iterate(self, max_iter=None):
...
+            If `None`, iterate indefinitely.
+
         Yields
         ------
         obj : Objects yielded by the streamer provided on init.
         '''
 
+        count = 0
         while True:
             for obj in self:
```
cjacoby: This isn't necessarily required for this PR, but I was thinking the other day that this `for obj in self` is a little ambiguous about what it is really doing with the new copies-of-streamers. I think I would prefer this line to be `for obj in self.iterate()`. While it is effectively the same, it is clearer what is going on under the hood.

bmcfee: Good call.  I think it's less an issue of clarity than removing a mutual recursion (call->cycle->call) in favor of the more direct (call->cycle->iterate).

PR #134, opened 2018-06-12
Small issues in the memory mapping example

hendriks73:
I think there are a couple of tiny issues in the memory mapping example.
Should be fixed with this PR.


bmcfee:
LGTM, thanks for this!

PR #138, opened 2018-11-23
fixed #127, target axis on buffer_stream

### LINKED ISSUE #127: buffer_stream option to append dimension
Sometimes your data has a batch index, sometimes it doesn't.  `buffer_stream` should let you specify whether to add an index dimension or not.

bmcfee:
This PR adds `axis` parameters to `buffer_stream` and `__stack_data`.  These can be used to prevent buffering from adding new axes to already buffered data, eg when batching up from batches.

I also removed an unused import.

This should be good to go, I think!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
     - pip install -e .[tests]
 
 script:
-    - py.test -v --cov-report= --cov=pescador tests/
+    - pytest
 
 after_success:
     - coveralls
--- a/.travis_dependencies.sh
+++ b/.travis_dependencies.sh
@@ -11,7 +11,7 @@ conda_create ()
     conda update -q conda
     conda config --add channels pypi
     conda info -a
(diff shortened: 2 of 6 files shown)


bmcfee:
The build is currently failing on 2.7 due to an unrelated issue with trapping warnings in the test suite. I suggest we ignore that for now, and fix it in a separate PR.

stefan-balke:
Hey Brian,
I just stumbled on this today and did not get from the docs whether it is useful for me to use it.
However, I saw you used it in `CREMA`. Could you drop 2 lines of docs here (when to use it and what the user gains from using it) while you are at it?

Thanks!

cjacoby:
I agree that it is would be useful to add documentation on this detail, as without it, it is initially quite unclear what exactly `buffer_stream` would do. I think it would also be useful to add an example in the docstring for `buffer_stream`, demonstrating a basic case.

Otherwise LGTM.

PR #150, opened 2019-07-18
Remove deprecated `Mux` and add Streamer conversion

bmcfee:
This PR implements #110 (automatic construction of streamer objects in mux construction).

To avoid doubling of effort, I also removed the old `Mux` class.  As long as test coverage stays good, I think we can merge this without CR.


bmcfee:
Coverage % fell due to removing a large chunk of covered code, but the new code has full coverage and tests pass.

Issue #129, opened 2018-03-01
A `streamable` decorator?


ejhumphrey:
**disclaimer**: this is a pretty simple proposal for some syntactic sugar.

I dunno about anyone else, but I find myself writing `stream = pescador.Streamer(foo, ...)` all over the place, where `foo` is some generator function I only ever want to use in pescador, for example...

```python
def random_samples(input_shape, k_classes, seed=None):
    rng = np.random.RandomState(seed)
    while True:
        yield dict(x=rng.uniform(size=input_shape), 
                   y=rng.randint(0, k_classes))
streamers = [pescador.Streamer(random_samples, input_shape, k_classes) 
             for _ in range(10000)]
mux = pescador.mux.StochasticMux(streamers, n_active=100, rate=100)
batches = pescador.Streamer(pescador.buffer_stream, mux, 32)
tuples = pescador.Streamer(pescador.tuples, batches, 'x', 'y')
``` 

Does anyone have any thoughts on introducing a `streamable` decorator to clean some of this up?

```python 
def streamable(func):
    def streamed(*args, **kwargs):
        return pescador.Streamer(func, *args, **kwargs)
    return streamed

@streamable
def random_samples(input_shape, k_classes, seed=None):
    rng = np.random.RandomState(seed)
    while True:
        yield rng.uniform(size=input_shape)

@streamable
def times_two(iterable):
    for x in iterable:
        yield x * 2

streamers = [random_samples(input_shape, k_classes) for _ in range(10000)]
mux = pescador.mux.StochasticMux(streamers, n_active=100, rate=100)
stream = times_two(mux)

for sample in stream.iterate(max_iter=10):
    # do something...
```

thoughts? concerns?

bmcfee:
I like the idea in principle.  Have you tried it?

cjacoby:
I like :+1:

ejhumphrey:
yea, i've been using it for the last couple days and i like it. I just copy-pasted out what i've been using above.

the *only* issue I think I've seen so far is that `Streamer`s aren't 100% interchangeable for generators (iirc) ... I've seen some things hiccup here and there, but didn't take notes and can't give an example. Adding a decorator can make it hard from context to determine the object type (generator vs Streamer) in separate code.

If I've got the right of this (which is definitely an if, I'm like 65% confident), then this might warrant a little bit more work to bring streamers just a little bit closer to generators. Y'all have any ideas on this?

bmcfee:
[...]

PR #151, opened 2019-07-18
Decorator interface

bmcfee:
This PR implements #129 

@ejhumphrey please give this a glance.  Tests pass and it behaves as expected.  I changed your implementation a little bit by using the `decorator` module.  This way, docstrings and call signatures are preserved:

```python
@pescador.streamable
def mystream(n):
    '''My bologna
    
    Docstring bonanza!
    '''
    for i in range(n):
        yield i
        
        
def streamable(func):
    def streamed(*args, **kwargs):
        return pescador.Streamer(func, *args, **kwargs)
    return streamed

@streamable
def mystream2(n):
    '''My bologna
    
    No docstring in sight.
    '''
    for i in range(n):
        yield i
```

results in
```python
[6]: mystream?
Signature: mystream(n)
Docstring:
My bologna

Docstring bonanza!
File:      ~/Notebooks/<ipython-input-5-0930abf6a0b9>
Type:      function
```

```python
[7]: mystream2?
Signature: mystream2(*args, **kwargs)
Docstring: <no docstring>
File:      ~/Notebooks/<ipython-input-5-0930abf6a0b9>
Type:      function
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ These use cases arise in the following common scenarios:
   Pescador makes it easy to interleave these sources while maintaining a small `working set`.
   Not all sources are simultaneously active, but Pescador manages the working set so you don't have to.
 
-- If loading data incurs substantial latency (e.g., due to accessing storage access
+- If loading data incurs substantial latency (e.g., due to accessing on-disk storage
   or pre-processing), this can be a problem.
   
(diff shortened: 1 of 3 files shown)


ejhumphrey:
lgtm. but wouldn't it be tuna over bologna for this lib? 🙃

Issue #93, opened 2017-05-05
Cache operations


ejhumphrey:
stochastic IIR filter for samples in a stream, providing sample reuse

ejhumphrey:
changed my mind, this is a better 2.0 milestone, there's gonna be some thought that goes into this design

bmcfee:
Had to unpack and explain this one today.  As I recall, here's what we actually had in mind:

`cache(stream, k, prob=0.5)`.  This maintains a cache of size `k` items.  With probability `prob`, it yields the next item from the stream and adds it to the cache (evicting the oldest item if `|cache|==k`).  With probability `1-prob` it yields a random item from the cache.

bmcfee:
Since nothing within pescador depends on this, can we move it up to a 2.1 milestone?

cjacoby:
👍

bmcfee:
Here's a rough cut of the cache map
```python
from pescador import PescadorError

def cache(stream, k, prob=0.5, random_state=None):
    '''Stochastic cache of a stream.
    
    With probability `prob`: yield a new item from `stream` and place it in the cache
    
    With probability `1-prob`: yield a previously seen item from the cache
    
    When the cache exceeds size `k`, a previously seen item is selected at random for eviction.
    
    Parameters
    ----------
    stream : iterable
        The stream from which to sample
        
    k : int > 0
        The size of the cache
        
    prob : float in [0, 1)
        The probability with which to select a new item
        
    random_state : None, int, or np.random.RandomState
        If int, random_state is the seed used by the random number generator;

        If RandomState instance, random_state is the random number generator;

        If None, the random number generator is the RandomState instance
        used by np.random.

    Yields
    ------
    data
        elements of `stream`
    '''
    if random_state is None:
        rng = np.random
    elif isinstance(random_state, int):
        rng = np.random.RandomState(seed=random_state)
    elif isinstance(random_state, np.random.RandomState):
        rng = random_state
    else:
        raise PescadorError('Invalid random_state={}'.format(random_state))

    cache = []
    
    while True:
        if len(cache) < k or rng.rand() >= prob:
            # We don't have enough data yet; grab an item and yield it
            cache.append(next(stream))
            yield cache[-1]
        else:
            yield rng.choice(cache)
            
        if len(cache) > k:
            cache.pop(rng.randint(low=0, high=k))
```

bmcfee:
Looking back over this, I think the comment above basically does everything we'd want?

[...]

PR #153, opened 2019-07-19
New maps

bmcfee:
This PR implements #93 -- iterator caching.

It also factors out the randomstate logic to a `util` function, since it now gets used in a few places.

### DIFF
--- a/pescador/maps.py
+++ b/pescador/maps.py
@@ -11,13 +11,15 @@ the function in a Streamer again.
     buffer_stream
     tuples
     keras_tuples
+    cache
 '''
 import numpy as np
 import six
 
+from .util import get_rng
 from .exceptions import DataError, PescadorError
 
-__all__ = ['buffer_stream', 'tuples', 'keras_tuples']
+__all__ = ['buffer_stream', 'tuples', 'keras_tuples', 'cache']
 
 
 def __stack_data(data, axis):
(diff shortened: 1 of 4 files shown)


bmcfee:
@ejhumphrey Can you chime in on this? I think we should not implement the validation function, so this is really just about the cache map.  I mostly want to know if this matches what you had in mind for this feature.

pescador/maps.py
```diff
@@ -236,3 +238,84 @@ def keras_tuples(stream, inputs=None, outputs=None):
             yield (x, y)
         except TypeError:
             raise DataError("Malformed data stream: {}".format(data))
+
+
+def cache(stream, n_cache, prob=0.5, random_state=None):
+    '''Stochastic stream caching.
+
+    With probability `prob`: yield a new item from `stream` and place it in the cache
```
ejhumphrey: this means the probability of the predicate happening, right? lower probs tend toward newer stuff, higher probs tend toward older stuff?

bmcfee: It's the probability of getting a new item rather than one you've seen before (as described below).

ejhumphrey: reading the rest of the docstring clarifies this ... perhaps it's enough for this docstring to speak to the intent of what's meant by caching here? something like "this is useful for effectively speeding up streaming by reusing in-memory datapoints, and should have a smoothing effect on gradients in online learning settings." the latter part is intuitive/hypothetical, but you know what I mean. maybe.

pescador/maps.py
```diff
@@ -236,3 +238,84 @@ def keras_tuples(stream, inputs=None, outputs=None):
...
+    ------
+    data
+        elements of `stream`
+    '''
+
+    if n_cache <= 0:
+        raise PescadorError('n_cache={} must be a positive integer'.format(n_cache))
+
+    if not 0 < prob <= 1:
+        raise PescadorError('prob={} must be a number between 0 and 1.'.format(prob))
```
ejhumphrey: maybe "in range (0, 1]"? it can be 1 but not zero, right?

bmcfee: :+1:

pescador/maps.py
```diff
@@ -236,3 +238,84 @@ def keras_tuples(stream, inputs=None, outputs=None):
...
+                item = next(stream)
+                yield item
+            except StopIteration:
+                break
+            data_cache.append(item)
+
+        else:
+            # Otherwise, the cache is now full.
+            # Pick a random index
+            idx = rng.randint(low=0, high=n_cache)
```
ejhumphrey: didn't we find once upon a time that generating random values one at a time is slow in large quantities? how often does this step get called, dya think? would it be premature to have a large array and shuffle it?

pescador/util.py
```diff
@@ -148,3 +150,34 @@ def batch_length(batch):
             raise PescadorError('Unequal field lengths')
 
     return n
+
+
+def get_rng(random_state):
```
ejhumphrey: 👌

ejhumphrey:
lgtm, coupla tiny things

pescador/maps.py
bmcfee: We did, but I think that problem is above the pay grade of this PR.  If we solve that, I'd prefer to do it at one level up of abstraction and across the entire library.

One potential solution to that would be to make a generator that pre-allocates a block of random numbers, yields them one at a time, and then replenishes when necessary.  If that can be done in numba, it should have pretty minimal overhead.

bmcfee:
Thanks for the CR!  Latest commit integrates your suggestions.  I want to punt on RNG efficiency until later.  I'll merge once this passes CI.

PR #155, opened 2019-08-01
Add support for Python 3.7, drop 2.7, 3.4 and 3.5

### LINKED ISSUE #149: Drop python 2 support
We really should stop trying to support python 2.  All the frameworks are now up to speed with python 3, and it will allow us to remove a good amount of shim code to keep legacy installs working.

Minimum python version should bump to at least 3.5, if not 3.6.  If we go to 3.6, we can have asynchronous generators (#30).

hugovk:
Fixes #149.


bmcfee:
Thanks @hugovk , but we still need to push out the 2.1 release before dropping support for py2.7.  It will be some time before we can get around to merging this.

hugovk:
Sure, that's fine. Are you planning on keeping Python 3.4 in 2.1? It reached [EOL on 2019-03-16](https://en.wikipedia.org/wiki/CPython#Version_history).

cjacoby:
Brian and I discussed this in-person a week and change ago, and concluded
that we should go to python>=3.6 so we can have async generators :-)

-C

bmcfee:
> Sure, that's fine. Are you planning on keeping Python 3.4 in 2.1? It reached [EOL on 2019-03-16](https://en.wikipedia.org/wiki/CPython#Version_history).

Not explicitly, but as long as we're still keeping legacy python for 2.1, I don't think there's any specific barrier to supporting 3.4 as well.  But as @cjacoby says, the plan for pesc>=3.0 is to require python 3.6+.

bmcfee:
This one conflicts with the other PR that just merged on 3.7 support.  I think we can close this out for now, and revisit deprecation later on when it becomes relevant again.

hugovk:
Okay!

It only took a moment to rebase the branch, so here it is :)

https://github.com/pescadores/pescador/compare/master...hugovk:rm-2?expand=1

PR #159, opened 2020-03-20
Add support for Python 3.8

hugovk:
Python 3.8 was released in October 2019.

And the Travis workaround for 3.7+ is no longer needed because Xenial is now the default.


bmcfee:
Thanks @hugovk for this!  Everything LGTM, and the coverage decrease seems to be unrelated.  I think we can merge this unless there are objections.

PR #164, opened 2024-01-24
2024 fixings

bmcfee:
Remove imp
Modernize setup script
Migrate from travis to gh actions


.github/workflows/ci.yml
```diff
@@ -0,0 +1,110 @@
...
+jobs:
+    test:
+        name: "Python ${{ matrix.python-version }} on ${{ matrix.os }}"
+        runs-on: ${{ matrix.os }}
+
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - os: ubuntu-latest
```
cjacoby, line 24: 3.7 is past end of life, do we need to support it?
https://devguide.python.org/versions/#versions

setup.cfg
```diff
@@ -1,2 +1,51 @@
...
+name = pescador
+version = attr: pescador.version.version
+description = Multiplex generators for incremental learning
+author = Pescador developers
+author_email = brian.mcfee@nyu.edu
+url = https://github.com/pescadores/pescador
+download_url = https://github.com/pescadores/pescador/releases
+long_description  =  Multiplex generators for incremental learning
+license = ISC
+python_requires = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
```
cjacoby, line 14: do we have known users using earlier versions than 3.7? They can always use an earlier version of pescador.

cjacoby (commented):
Aside from the checks not passing, and the <python=3.7 questions, lgtm. Also 👋 hey :)

.github/workflows/ci.yml
bmcfee: My plan for this PR was to only change metadata and actions, but leave the package unchanged.  That way we could put out a post-release that doesn't impact any dependencies.

It probably would be simpler to just modernize everything, but I'm pretty limited on mental bandwidth for the time being (currently also maintaining a new human fork).  Any interest/available to chip in on this PR?

cjacoby: I've had in the back of my mind to modernize some things here, because I had some troubles a little while ago with the old joblib dependency. I have interest but limited availability due to my own fork with a recent 20 month release. 

What if we basically merge this one as is, and I'll try to make smaller PRs with little improvements when I can?

(PS Congrats!)

bmcfee: I'd like to at least get the environment issues squared away first so that the tests run (even if they don't pass, but they should).  I think it mainly involves spending a few minutes to sort out what versions of everything are actually available on conda-forge, and probably removing the upper version pin on pytest.

Otherwise, yeah, we should do some modernization here, but agree that should be a separate PR.

(PPS Thanks, and congrats on yours too!)

bmcfee: ... I now see what you mean about the joblib issue.  That will take some doing.

bmcfee:
Ok, I think this is finally good to go.

There are some other modernization steps needed now (eg removing six dependency), but at least the CI runs and we're in the green.

Issue #148, opened 2019-07-16
Alternate distributions for StochasticMux


bmcfee:
I've been doing a bit of analysis of pescador's sampling behavior, and the Poisson distribution is a bit unwieldy.  However, for a specific choice of parameters, drawing `n_samples_to_stream` from a binomial distribution leads to a clean solution in terms of total replacement rate of streamers.

More generally, I think it would be useful to extend `StochasticMux` to have a `dist=` parameter that works as follows:

- `constant`: each streamer gets exactly (at most) `rate` samples.
- `binomial` (proposed new default): each streamer gets a binomial draw from `Binomial(rate, 1-1/n_active)`
- `poisson`: current behavior, `n_samples ~ 1 + Poisson(rate)`

bmcfee:
Quick summary of the analysis: say we have uniform weights over `n_active` active streamers.  Each streamer has a rate limit that's sampled from a binomial distribution 

`n_samples ~ Binomial(rate * n_active / (n_active - 1), (n_active - 1) / n_active)`.

- The expectation of `n_samples` is `rate`
- Pick any active streamer, and let `M` denote the (random) number of samples we need to draw from the mux before that streamer gets replaced.  This has expected value `E[M] = rate * n_active`.
- The variance is `Var[M] = rate * n_active**2`.

These expectations match exactly the case where the number of samples is deterministic: `n_samples = rate`.  The benefit of randomizing `n_samples` is that you don't get a mass extinction event every `rate + K` samples; the replacements are less concentrated.

In exhaustive mode, we want to know how long until we hit all the streamers.  We can think of this as each streamer slot being replaced `n_streams/n_active - 1` times, and the expected time cooks down to `rate * (n_streams - n_active) / n_active` (exactly what you'd hope for).

Finally, we'd want to know what the chance of a streamer *not* being replaced is.  This can be calculated numerically using the negative binomial survival function (which is ugly AF), but we could also be lazy and use some concentration inequalities.  If we think of `rate * n_active` as the average time of an 'episode', then we could ask what the chance of a streamer lasting some `b>=1` episodes longer than average.  This probability is upper-bounded by `1/(b**2 * rate)`; so with high probability, every streamer will be replaced quickly and the distribution should not become unbalanced.  (Corollary: the approximation for time to exhaustion above should be pretty good.)

lostanlen:
Cool!

So is pescador going to be renamed to pascalador then?*


[...]

Issue #132, opened 2018-03-15
on StochasticMux rates, scalloping loss contours


ejhumphrey:
I've just done something rather foolish (so this is documented user error), but wanted to share (a) the experience in case anyone else steps on it, and (b) an idea for improving upon StochasticMux to (potentially) help avoid this in the future.

First, I lost track of how many "unique" samples were in each bag, and the rate was too high. This can lead to "scalloped" training contours and _really_ slow learning ... here's a diagram for accuracy over iteration.

![image](https://user-images.githubusercontent.com/926379/37485429-cf401120-2861-11e8-9377-7bf301e0f71c.png)

What's happening is that the data stream isn't shuffling through the data fast enough, and overfitting a relatively small number of datapoints at a time. As the underlying muxes shuffle through the data, you get sharp drops in training accuracy when new streams are activated (more on this in a second). Then, because the model is sitting on the same data too long, it's not learning all that well, leading to a significant gap in performance with the validation data. 

tl;dr: turn your rate down.

Second, and perhaps more interesting, the sharp drops occur because the streams are all getting refreshed at roughly the same time. While the StochasticMux should sort this out over time (which it looks like is happening), it takes a while for this staggering to occur. One option would be to treat the first round of stream activations differently, instead drawing uniformly from, e.g. [1, rate]. I ... think .. this should be enough to decorrelate the various streamers from the outset.

I also *think* that change would mostly be confined to [mux.py:L754](https://github.com/pescadores/pescador/blob/master/pescador/mux.py#L754), with a little bit of logic in the __init__ method to indicate that it's the first round. 

thoughts?

bmcfee:
> Second, and perhaps more interesting, the sharp drops occur because the streams are all getting refreshed at roughly the same time.

Do you still have the code from this example?  I'd be curious to know what the parameters `(rate, n_streamers, n_active)` were (as well as batch size for your training), and how that squares with the theoretical analysis we have for replacement rates.

bmcfee:
>  One option would be to treat the first round of stream activations differently, instead drawing uniformly from, e.g. [1, rate]. I ... think .. this should be enough to decorrelate the various streamers from the outset.

That's an interesting idea -- essentially adding a random offset to each streamer slot's expected replacement time.  I don't think it should be default behavior, but it's a nice optimization.

bmcfee:
Alternately, I wonder if using a stream cache could help smooth that out?

PR #171, opened 2024-02-01
Revised distributions for stochasticmux

bmcfee:
This PR implements several changes described in #148 

- New distributions for stochasticmux: `const` and `binomial` (new default)
- Adjusted the poisson mode so that the expected value is actually `rate` and not `rate+1`

I've also relaxed the uniform convergence unit tests.  A p-value of >=0.95 was probably overkill for the sample size we were drawing, and I've reduced it to 0.5.  Strangely, poisson was giving me the most trouble here, while const and binomial were behaving better.  It's probably an artifact of setting `rate=2` in the test.


bmcfee:
There's a bit of weirdness here in the initialization with binomial mode.  (Will come back to this later...)

The binomials are parametrized by `Bin((rate-1)/(1-p), 1-p)` where `p` is the probability of selecting the streamer from the active set.  The reason for this dependence is subtle, but it ensures that the replacement times for streamers don't concentrate too much; a streamer gets replaced on average every `rate * N_active` samples, with variance like `rate * N_active * (N_active-1)`.  (At least, assuming uniform weights on the streamers.)

The weirdness arises when we initialize streamers on at a time: the weight distribution is not fully known until the first batch of active streamers are fully initialized, so the calculations for `p` above are generally going to be wrong.  In the extreme case, the first active streamer will use the poisson approximation (since `p=1` when there are no other streamers active yet); the second streamer will have `bin((rate-1)/0.5, 0.5)` (assuming uniform weights again), the third will have `bin((rate-1)/0.666, 0.666)`, and so on.  In all cases, the expected values will be the same, but the first few streamers will have higher rate variance than the later ones.  Specifically, the variance sequence will look like (again, assuming uniform streamer weights) of `(rate-1)/n` (for `n=1,2, ..., n_active`).

Now, we could hack around this by pre-determining the weights so that everything is primed properly.  However, I think it might actually be beneficial to leave it as is because it injects more randomness in the rate distributions early on, which ought to have a comparable effect to having random offsets in a burn-in phase as @ejhumphrey suggested in #132 .

It's a bit weird, but given the potentially dynamic nature of the active stream distribution (especially in exhaustive mode), it won't be possible to always ensure that the rate distribution for a streamer is "correct" over time.  The best we can do is sample the rate value according to whatever the distribution will be at the time the streamer is activated.

bmcfee:
Having slept on it, i think a better solution here is to initialize the active set weights by a random draw from the weights array instead of with zeros. This won't have any effect on const or poisson, but it will put the binomial mode in a less quirky position at initialization time.

bmcfee:
@cjacoby :wave: I know it's been a gajillion years, but do you have any interest in looking this over?  I think it's basically good to go, but it does have some kinda breaky behavior relative to older versions that I'd like to get another set of eyes on.

Quick TLDR is summarized in https://github.com/pescadores/pescador/issues/148#issuecomment-1919834323

pescador/mux.py
```diff
@@ -335,7 +336,12 @@ def __init__(
                 Run every selected stream once to exhaustion.
 
         prune_empty_streams : bool
-            Disable streamers that produce no data. See `BaseMux`
+            Disable streamers that produce no data.
+
+        dist : ["constant", "binomial", "poisson"]
+            Distribution governing the (maximum) number of samples taken
+            from an active streamer.
+            In each case, the expected number of samples will be `rate`.
```
cjacoby, line 344: I think for best user experience, it would be nice to have a quick summary of (as a consumer of the library) when you might want to use each of these distributions.

Or, perhaps a tl;dr, with a link back to the analysis in the issue for in depth understanding.

cjacoby (changes requested):
lgtm other than I think a small comment improvement would improve it (for me in 6mo to a year when I come back and can't remember what this is about).

pescador/mux.py
bmcfee: That's a good point. I think what I'll do is write up a section for the docs that goes over this in detail. Will try to get that done some time in the next week.

bmcfee:
Ok, doc section is added and back-link is included.  I tried to clean it up a bit from my original notebook (4 years ago!) and put in some expository text.  Hopefully it makes sense?

cjacoby (approved):

